PR #1, opened 2017-06-27
Url Module: Bootstrap the repository and add the first module

youknowriad:
This PR tries to bootstrap the repository using a simple Lerna config.
This PR is highly inspired by how the [Jest](https://github.com/facebook/jest) Repository works.
The idea is trying to build our workflow around a simple module breaking things up into several PRs. 
For this initial PR, I'm setting up the `build` script. (Trying to keep it as small as possible)
Follow-up PRs should think about linting, testing, watching...

Included in this PR:

 - The `url` module as as an example module. The example function here is actually extracted from The Gutenberg repository.
 - A build script allowing us to build all the modules with the same script (avoid duplicating the logic in all submodules)
 - The babel config is the one used for Gutenberg

Not included yet:

 - The browser build should automatically generate the `wp.url` global variable.

**Testing instructions**

```bash
npm install --global lerna@^2.0.0-beta.0
npm install
lerna bootstrap
npm run build
```


lerna.json
```diff
@@ -0,0 +1,7 @@
+{
+  "lerna": "2.0.0-rc.5",
+  "packages": [
+    "packages/*"
+  ],
+  "version": "0.0.0"
+}
```
rmccue: Should we use tabs in JSON config too? 🤔

packages/url/package.json
```diff
@@ -0,0 +1,18 @@
...
+  "name": "@wordpress/url",
+  "version": "0.0.1",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/WordPress/packages.git"
+  },
+  "description": "WordPress URL utilities",
+  "main": "build/index.js",
+	"module": "build-module/index.js",
+	"browser": "build-browser/index.js",
```
rmccue: Mixed indentation.

scripts/build.js
```diff
@@ -0,0 +1,151 @@
...
+	module: 'build-module',
+	browser: 'build-browser',
+};
+const OK = chalk.reset.inverse.bold.green( ' DONE ' );
+
+/**
+ * Babel Configuration
+ */
+const babelDefaultConfig = JSON.parse(
+  fs.readFileSync( path.resolve( __dirname, '..', '.babelrc' ), 'utf8' )
```
rmccue: Mixed indentation.

scripts/watch.js
```diff
@@ -0,0 +1,69 @@
...
+const BUILD_CMD = `node ${path.resolve(__dirname, './build.js')}`;
+
+let filesToBuild = new Map();
+
+const exists = filename => {
+  try {
+    return fs.statSync(filename).isFile();
+  } catch (e) {}
+  return false;
+};
```
rmccue: Should probably tabify this file too.

lerna.json
ntwb: Yes, I'll add an `.editorconfig` shortly, WordPress is switching back to tabs for `.json` files:

See  [#WP40946](https://core.trac.wordpress.org/ticket/40946) and https://github.com/WordPress/gutenberg/pull/1067

youknowriad: This is auto-generated by lerna, I wonder if it forces spacing indentation if I replace with tabs. I'll test.

youknowriad:
I'll fix those indentation issues, hopefully we could get eslint soon :)

ntwb:
@youknowriad I'll have an ESLint PR soon 😄

youknowriad:
I'd appreciate some 👀  and a ✅ to move forward with the other steps (testing, watching...)

.babelrc
```diff
@@ -0,0 +1,20 @@
...
+		[ "env", {
+			"targets": {
+				"browsers": [
+					"last 2 Chrome versions",
+					"last 2 Firefox versions",
+					"last 2 Safari versions",
+					"last 2 iOS versions",
+					"last 1 Android version",
+					"last 1 ChromeAndroid version",
+					"ie 11"
```
atimmer: Did you omit the last 2 edge versions on purpose?

package.json
```diff
@@ -0,0 +1,16 @@
...
+    "babel-plugin-transform-object-rest-spread": "^6.23.0",
+    "babel-plugin-transform-runtime": "^6.23.0",
+    "babel-preset-env": "^1.5.2",
+    "chalk": "^1.1.3",
+    "glob": "^7.1.2",
+    "lerna": "^2.0.0-rc.5",
+    "mkdirp": "^0.5.1"
+  },
+  "scripts": {
+    "build-clean": "rm -rf ./packages/*/build ./packages/*/build-es5",
```
atimmer: This won't work on windows. Is that a concern? If so, we could use `rimraf`.

atimmer:
I have commented in two locations. I have also tested this locally, for me this PR is ✅. Concerns could be addressed in new PRs.

.babelrc
youknowriad: I've just copy/pasted from Gutenberg, we missed Edge :)

package.json
youknowriad: Good catch, let's not leave Windows users behind :)

youknowriad:
Thanks for the review, Feedback addressed.

gziolo:
Cool, lerna seems like the best way to go here 👍

PR #9, opened 2017-07-03
Test: Setup unit testing using Jest

youknowriad:
This PR tries to setup unit testing using the `Jest` Framework and could serve as a starting point of the discussion to choose our Test Runner. 

Jest has solid arguments for it: matchers, mocks, snapshots, an AWESOME cli, coverage webpack and babel integration...), easily configurable

We are using mocha and webpack in Gutenberg but we have some issues (running and building target tests is one of them), Jest resolves them.

I kept it simple, just a unique `npm` script running `jest`.

If you want to experience some niceties of the Jest cli, try running  `npm test -- --watch` or `npm test -- --coverage`

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,12 +6,14 @@
     "babel-preset-env": "^1.5.2",
     "chalk": "^1.1.3",
     "glob": "^7.1.2",
+    "jest": "^20.0.4",
     "lerna": "^2.0.0-rc.5",
     "mkdirp": "^0.5.1",
     "rimraf": "^2.6.1"
   },
   "scripts": {
     "build-clean": "rimraf ./packages/*/build ./packages/*/build-browser ./packages/*/build-module",
-    "build": "node ./scripts/build.js"
+    "build": "node ./scripts/build.js",
+    "test": "jest"
   }
 }
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
   "scripts": {
     "build-clean": "rimraf ./packages/*/build ./packages/*/build-browser ./packages/*/build-module",
     "build": "node ./scripts/build.js",
-    "test": "jest"
+    "test": "jest",
+    "watch": "jest --watch"
   }
 }
--- a/packages/url/src/test/index.test.js
+++ b/packages/url/src/test/index.test.js
@@ -1,0 +1,20 @@
+/**
+ * Internal Dependencies
+ */
+import { addQueryArgs } from '../';
+
+describe( 'addQueryArgs', () => {
+	test( 'should append args to an URL without query string', () => {
+		const url = 'https://andalouses.com/beach';
+		const args = { sun: 'true', sand: 'false' };
+
+		expect( addQueryArgs( url, args ) ).toBe( 'https://andalouses.com/beach?sun=true&sand=false' );
+	} );
+
+	test( 'should append args to an URL with query string', () => {
+		const url = 'https://andalouses.com/beach?night=false';
+		const args = { sun: 'true', sand: 'false' };
+
+		expect( addQueryArgs( url, args ) ).toBe( 'https://andalouses.com/beach?night=false&sun=true&sand=false' );
+	} );
+} );
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@ node_modules/
 /packages/*/build/
 /packages/*/build-module/
 /packages/*/build-browser/
+
+coverage/


package.json
```diff
@@ -6,12 +6,14 @@
...
+    "jest": "^20.0.4",
     "lerna": "^2.0.0-rc.5",
     "mkdirp": "^0.5.1",
     "rimraf": "^2.6.1"
   },
   "scripts": {
     "build-clean": "rimraf ./packages/*/build ./packages/*/build-browser ./packages/*/build-module",
-    "build": "node ./scripts/build.js"
+    "build": "node ./scripts/build.js",
+    "test": "jest"
```
ntwb: Speaking of the awesomeness of Jest watch, how about a watch task here below this line:

> `"watch": "jest --watch"`

youknowriad: Actually, I wanted to make this default in `npm test` but I was wondering if `jest` is smart enough to disable watching in non interactive CLI (travis)

gziolo: https://facebook.github.io/jest/docs/en/cli.html#onlychanged - is it what you are looking for? :) It is executed in watch mode by default and runs test only on modified files.

gziolo:
Thanks for opening this PR. I think it is a very good start which shows how easy it is to setup Jest :) Personally I echo what @ntwb proposed regarding to having its own command for watch mode.  In my another PR for Gutenberg project I also included a command which generates code coverage (https://github.com/WordPress/gutenberg/pull/1382). In Calypso project I proposed also a command for CI, because CircleCI that we use requires a few tweaks (https://github.com/Automattic/wp-calypso/pull/14364). I think it's much easier for new developers to check `package.json` to see possible commands and play with them, rather than looking into Jest documentation. So we can have a wide variety of the commands like:
- `npm test`
- `npm test:ci`
- `npm test:coverage`
- `npm test:watch`

We can add them later when we agree that they are useful.

package.json
ntwb: Hmmm, not sure _watch_ should be included in the default `npm test` task to be honest 🤔 

At the moment I'm refreshing the `eslint` branch for #7 and swapping a couple of things around, and this is what I've got at the moment:

```json
  "scripts": {
    "build-clean": "rimraf ./packages/*/build ./packages/*/build-browser ./packages/*/build-module",
    "lint": "eslint . --ignore-path .gitignore",
    "pretest": "npm run lint",
    "test": "jest",
    "watch": "jest --watch"
  }
```

I doubt `pretest` would run the ESLint task if Jest is in _watch_ mode...

youknowriad: After reconsidering, I think you're right in a "library" repository, the watch command should watch and run the unit tests, this looks good to me. I'll update.

youknowriad:
I'll wait a bit here until the meeting (which I'll probably miss, I'll be between two flights). I hope to see this discussed there and we could merge after.

adamsilverstein:
Looks good! I also like the idea of adding the npm commands in package.json.

package.json
gziolo: > I doubt pretest would run the ESLint task if Jest is in watch mode...

You can modify `scripts` to run lint when initializing watch mode with this change:

```js
 "scripts": {
    "build-clean": "rimraf ./packages/*/build ./packages/*/build-browser ./packages/*/build-module",
    "lint": "eslint . --ignore-path .gitignore",
    "pretest": "npm run lint",
    "test": "jest",
    "watch": "npm test -- --watch"
  }
```

I'm not sure it is possible to re-trigger eslint on every watch mode run, but it should be okey.

gziolo:
Also we should consider using Multi-Project Runner. It is described in this blog post announcement: http://facebook.github.io/jest/blog/2017/05/06/jest-20-delightful-testing-multi-project-runner.html#multi-project-runner-configuration-overhaul. We probably don't need it at the moment, but as this repository will grow, it might be very useful to have possibility to add customized configuration for the specific package. On the contrary Jest repository uses one global config that works with all packages they publish with Lerna. See https://github.com/facebook/jest/blob/master/package.json#L92. I'm rather posting it here to have it somewhere on the roadmap 😄

youknowriad:
I'm not against adding new scripts but I'd like us to wait and add them later when necessary. I expect us adding a lot of npm scripts over time. Thus, we should also consider keeping it at a manageable size.

packages/url/src/test/index.test.js
```diff
@@ -0,0 +1,20 @@
+/**
```
gziolo: At the moment we use a default glob base pattern to match test files which looks this way:
```js
[ '**/__tests__/**/*.js?(x)', '**/?(*.)(spec|test).js?(x)' ]
```

Jest repository uses the following highly customized mix:
```js
"testPathIgnorePatterns": [
    "/node_modules/",
    "/examples/",
    "/integration_tests/.*/__tests__",
    "\\.snap$",
    "/packages/.*/build",
    "/packages/.*/build-es5"
],
"testMatch": [
    "**/*.test.js"
]
```

In Calypso we are targeting:
```js
"testMatch": [
 	"**/bin/**/integration/*.js",
 	"**/client/**/integration/*.js",
 	"**/server/**/integration/*.js"
 ],
```

and

```js
"testMatch": [ "**/server/**/test/*.js?(x)" ],
```

I don't know what is the most popular option in WordPress world, but we should rather follow what people use :)

ntwb: The `/**` the above comment is on is for a file header comment @gziolo 

We can add these paths in `package.json` after merging this PR.

As to what's popular in the WordPress world for this, not much prior art, so "best practices" is what we should aim for

ntwb:
Agreed on adding new ''npm scripts'' only as required, lets try to keep them under control (and also not overly long either, under 80 characters is 👌 )

packages/url/src/test/index.test.js
gziolo: > The /** inthe above comment is on is for a file header comment @gziolo, not Jest 😏

I put this comment in random place, sorry for confusion. We use those comments for dependencies in Calypso. I personally hope we don't inherit them to WordPress world :D

gziolo:
Jest by default uses `jsdom` as test environment. See http://facebook.github.io/jest/docs/en/configuration.html#testenvironment-string. What it means is it tries to fake that test is executed in browser. It is possible to use also `node` env which is much faster. See this benchmark: https://medium.com/@vitaliypotapov/thanks-for-suggestion-jest-is-really-30-faster-with-env-node-option-the-chart-4309ba8aba81. We should keep an eye on the type of packages that are introduced. If majority of them is not going to refer `window` object we should change test environment to `node`. It is still possible to change test env per test suite with comment at the very beginning of the file.

packages/url/src/test/index.test.js
ntwb: There's no place for random comments on unrelated code in a pull request, they should not never be added.

Comments such as this original comment should be added as a comment on the PR or a new issue created but never random comments on random lines of code of a pull request, its beyond confusing IMHO.

omarreiss:
👍 for Jest. 👍 for npm scripts for generic workflow commands that everyone developing in this repo will need to use.

packages/url/src/test/index.test.js
adamsilverstein: Looks good, I'd love to start using Jest here. We may have some QUnit tests to move over if we move any of the files we have tests for into modules.

PR #10, opened 2017-07-07
Adding WordCounter package

ryanwelcher:
This a first pass at adding word-count.js as a package.

There is a dependency on the `window.l10n` object which I have not updated and causes the shortcode related tests to fail. Those tests are in place but commented out for now.

Any comments/recommendations are welcome


youknowriad:
I know this is an existing package, but why is it an Object constructor instead of a simple function? and why it needs this dependency to work?

I would prefer a simpler API like this:

```js
const count = wordCount( text, settings );
```

ryanwelcher:
@youknowriad I've created a simpler API as per your suggestion. Please have a look and let me know if you have any suggestions.

The `i10l` dependency is needed to process the shortcodes and contains the regexs

nylen:
I believe @iseulde worked on the original code in core and could provide input on why the existing API is structured as a class.

iseulde:
I don't care that much tbh. But maybe remove the class entirely then. Creating a new instance for every count seems odd.

ryanwelcher:
I'm planning on refactoring this to make more sense in this context. Hoping to have to done in the next couple of days. Thanks for the feedback!

ryanwelcher:
I've pushed changes to use this as a function.

packages/wordcount/src/index.js
```diff
@@ -0,0 +1,193 @@
+/**
+ * Word counting utility
+ *
+ * Counts the number of words (or other specified type) in the specified text.
+ *
+ * @summary  Count the number of elements in a text.
+ *
+ * @since    2.6.0
+ * @memberof wp.utils.wordcounter
```
youknowriad: Should we drop this line, I guess it will be different now and I wonder if we'd have a global at all.

packages/wordcount/src/index.js
```diff
@@ -0,0 +1,193 @@
...
+	// Apply provided settings to object settings.
+	if ( settings ) {
+		for ( let key in settings ) {
+
+			// Only apply valid settings.
+			if ( settings.hasOwnProperty( key ) ) {
+				defaultSettings[ key ] = settings[ key ];
+			}
+		}
+	}
```
youknowriad: Can't we replace this loop with something like:

```js
settings = { ...defaultSettings, ...settings );
```

and use `settings` instead of `defaultSettings` later?

nylen:
> Should we drop this line, I guess it will be different now and I wonder if we'd have a global at all.
 
This is something that needs to be planned and tested pretty carefully: How will plugins consume this code?

Will we replace or remove the current global inside `wp.util`? If so, we cannot change the API that plugins are already using.

packages/wordcount/src/index.js
nylen: It should have a different name (`settings` and `userSettings` perhaps) but it doesn't matter because the object is recreated every time.

aduth: Worth noting that object spread / `Object.assign` is shallow, which doesn't seem to be a problem at the moment, but could be if more nested object values (like `options.l10n`) are added in the future, since it would not be possible to specify some but not all of the nested values (i.e. rely on a nested merge behavior).

packages/wordcount/src/index.js
```diff
@@ -0,0 +1,193 @@
...
+ * @param {String} settings.l10n.type                         Optional. Method of finding words to count.
+ * @param {Array}  settings.l10n.shortcodes                   Optional. Array of shortcodes that should be removed
+ *                                                            from the text.
+ *
+ * @return {Number} The number of items counted.
+ */
+export function count( text, type, settings ) {
+	let count = 0;
+	let shortcodes;
+	const defaultSettings  = {
```
aduth: If there's nothing about these defaults that depends on the function arguments, can we move it to the top of the file so we're not creating new references / executing RegExp `Array#join` on every function call?

ryanwelcher:
Hi all - I've made updates based on the notes above. I've also added lodash as a dependency and am using `_.extend` instead of the spread operator to manage the settings. I believe this addresses the shallow copy concerns. Please me know if that is overkill and I'm happy to revert to using `settings = { ...defaultSettings, ...userSettings }`

packages/wordcount/src/index.js
```diff
@@ -31,123 +36,108 @@
...
 		}
 	}
 
-	shortcodes = this.settings.l10n.shortcodes;
+	shortcodes = defaultSettings.l10n.shortcodes || {};
 
 	// If there are any localization shortcodes, add this as type in the settings.
 	if ( shortcodes && shortcodes.length ) {
-		this.settings.shortcodesRegExp = new RegExp( '\\[\\/?(?:' + shortcodes.join( '|' ) + ')[^\\]]*?\\]', 'g' );
+		defaultSettings.shortcodesRegExp = new RegExp( '\\[\\/?(?:' + shortcodes.join( '|' ) + ')[^\\]]*?\\]', 'g' );
```
sabitertan: This can throw error if shortcodes set to a empty object.

packages/wordcount/src/index.js
```diff
@@ -0,0 +1,187 @@
...
+
+			// Remove unwanted characters.
+			text = text.replace( settings.removeRegExp, '' );
+		} else {
+
+			// Convert HTML Entities to "a".
+			text = text.replace( settings.HTMLEntityRegExp, 'a' );
+
+			// Remove surrogate points.
+			text = text.replace( settings.astralRegExp, 'a' );
```
omarreiss: Then finally this would be `transposeAstralsToCountableChars`.

nylen:
Before we get too deep into refactoring, we need to think about how the code in this repository will actually be used...  Quoting from [above](https://github.com/WordPress/packages/pull/10#issuecomment-317699268)

> How will plugins consume this code?

> Will we replace or remove the current global inside `wp.util`? If so, we cannot change the API that plugins are already using.

omarreiss:
@nylen we can easily keep the API inside WordPress intact. This is how I would see this implemented in core (docs left out for convenience):

```JS
import { count } from '@wordpress/count';

function WordCounter( settings ) {
  this.settings = settings || {};
}

WordCounter.prototype.count = function( text, type ) {
  return count( text, type, this.settings );
}

window.wp = window.wp || {};
window.wp.utils = window.wp.utils || {};
window.wp.utils.WordCounter = WordCounter;
```

Main difference is that the default settings are now delegated to the `count` package. For BC sake, we could decide to keep the default settings in core as well but that is a different discussion and one that should probably be held in the trac ticket to start using the module.

With regard to my refactoring suggestions, I don't think there is any reason to hold off on applying those, as they don't really affect the behavior in any way or change any API. They just make the code more reusable and easier to comprehend.

ryanwelcher:
@nylen, @omarreiss for what it's worth, core seems to use the current implementation without any settings being passed - which simplifies the dependency/BC issue a bit.
See - https://core.trac.wordpress.org/browser/tags/4.8/src/wp-admin/js/post.js#L1267

@omarreiss thanks for your feedback. I'll work on implementing it shortly.

nylen:
It doesn't look like any plugins are passing custom settings to `WordCounter` either.  Here are the plugins making use of `WordCounter`:

```
$ ./summarize-scan.php scans/WordCounter.txt
3 matching plugins
Matches  Plugin                     Active installs
=======  ======                     ===============
      1  nelio-content                       4,000+
      6  wpglobus                           10,000+
      1  visual-term-description-editor      7,000+
```

ryanwelcher:
@omarreiss I've refactored this package based on your suggestions. Any feedback is appreciated!

ryanwelcher:
A little more refinement on the approach to abstract files out and use a common settings object.

packages/wordcount/src/index.js
```diff
@@ -0,0 +1,119 @@
+import _ from 'lodash';
```
youknowriad: Should we import directly the dependencies we need: `import { extend } from 'lodash'`?
And maybe we can use https://github.com/lodash/babel-plugin-lodash to allo tree-shaking

ryanwelcher:
@youknowriad I've refactored to use `_.flow`, only pull the items we need out of lodash, and added a README. There are already existing Jest tests that I converted from the existing core ones - did you want anything else tested?

packages/wordcount/package.json
```diff
@@ -0,0 +1,17 @@
...
+  "name": "@wordpress/wordcount",
+  "version": "0.0.1",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/WordPress/packages.git"
+  },
+  "description": "WordPress Word Count Utility",
+  "main": "build/index.js",
+  "module": "build-module/index.js",
+  "browser": "build-browser/index.js",
```
youknowriad: We removed the browser build for now

packages/wordcount/src/index.js
```diff
@@ -0,0 +1,112 @@
...
+
+/**
+ * Private function to manage the settings.
+ *
+ * @param type
+ * @param userSettings
+ * @returns {void|Object|*}
+ */
+function loadSettings( type, userSettings ) {
+	let settings = extend( defaultSettings, userSettings  );
```
youknowriad: Shoud this be `const` instead of `let`?

youknowriad: I guess we should add an ESlint rule to make these consistent.

packages/wordcount/src/index.js
```diff
@@ -0,0 +1,112 @@
...
+/**
+ * Count some words.
+ *
+ * @param {String} text
+ * @param {String} type
+ * @param {Object}  userSettings
+ *
+ * @returns {Number}
+ */
+export const WordCounter = {
```
youknowriad: Should we change the API like proposed before  https://github.com/WordPress/packages/pull/10#issuecomment-313775265

Unless I've missed a comment, it seems we have a consensus to use the functional approach and ensure the Backwards compatibility when importing this in WP (if needed)

gziolo:
@ryanwelcher, it looks to be abandoned. Do you plan to continue work on this PR? It needs to be rebased to apply the latest changes. I guess some of the changes in `package.json` added in this PR will have to be removed.

packages/wordcount/README.md
```diff
@@ -0,0 +1,17 @@
...
+Install the module
+
+```bash
+npm install @wordpress/wordcount@next --save
+```
+
+## Usage
+
+```JS
+import {WordCounter} from '@wordpress/wordcount';
```
youknowriad: Should we update the docs to match the last API?

packages/wordcount/README.md
```diff
@@ -0,0 +1,17 @@
+# @wordpress/wordcount
+
+A utility to count words
+
+## Installation
+
+Install the module
+
+```bash
+npm install @wordpress/wordcount@next --save
```
youknowriad: I believe we can remove `@next` since we'll probably release a stable version post merge.

packages/wordcount/README.md
```diff
@@ -0,0 +1,17 @@
...
+```
+
+## Usage
+
+```JS
+import {WordCounter} from '@wordpress/wordcount';
+const count = WordCounter.count( 'Words to count', 'words', {});
```
youknowriad: Minor: Maybe addding some documentation explaining the arguments and the important options we can pass.

packages/wordcount/src/index.js
```diff
@@ -0,0 +1,102 @@
...
+
+
+
+/**
+ * Private function to manage the settings.
+ *
+ * @param type
+ * @param userSettings
+ * @returns {void|Object|*}
+ */
```
youknowriad: I believe the last recommendation, we use `@return` instead of return. And we should type each param and add a description. Something like

```
* @param {string}           type              short description of the type arg
* @param {Object}         userSettings short description of the user settings 

* @returns {void|Object|*} description of the returned value
```

packages/wordcount/src/index.js
```diff
@@ -0,0 +1,102 @@
...
+import stripTags from './stripTags';
+import transposeAstralsToCountableChar from './transposeAstralsToCountableChar';
+import stripHTMLEntities from './stripHTMLEntities';
+import stripConnectors from './stripConnectors';
+import stripRemovables from './stripRemovables';
+import stripHTMLComments from './stripHTMLComments';
+import stripShortcodes from './stripShortcodes';
+import stripSpaces from './stripSpaces';
+import transposeHTMLEntitiesToCountableChars from './transposeHTMLEntitiesToCountableChars';
+
```
youknowriad: I belive multiple empty lines are not allowed in our eslint rules (which will be integrated to this repository soon I guess)

packages/wordcount/src/index.js
```diff
@@ -0,0 +1,102 @@
...
+/**
+ * Match the regex for the type 'words'
+ * @param text
+ * @param regex
+ * @returns {Array|{index: number, input: string}}
+ */
+function matchWords( text, regex, settings ) {
+	text = flow(
+		stripTags.bind( this, settings ),
```
youknowriad: `.bind` calls are useless now that we don't use a class

ryanwelcher:
@youknowriad I've implemented the requested changes. I've left the `.bind()` in as we need it to get the `settings` parameter into each of the functions.

Please let me know if there is anything else. Thanks for your help and really looking forward to moving this ahead!

packages/wordcount/README.md
```diff
@@ -0,0 +1,26 @@
+# @wordpress/wordcount
+
+A utility to count words
+
+## Installation
+
+Install the module
+
+```bash
+npm install @wordpress/wordcount--save
```
youknowriad: missing space

packages/wordcount/README.md
```diff
@@ -0,0 +1,26 @@
...
+count( text, type, userSettings )
+````
+count accepts three parameters:
+1. text: A string containing the words/characters to be counted.
+2. type: A string that represents the type of count. The current implementation accepts the strings 'words', 'characters_excluding_spaces', or 'characters_including_spaces'.
+3. userSettings: An object that contains the list of regular expressions that will be used to count. See defaultSettings.js for the defaults.
+
+## Usage
+```JS
+import {count} from '@wordpress/wordcount';
```
youknowriad: missing spaces `import { count } from '@wordpress/wordcount';`

packages/wordcount/src/stripConnectors.js
```diff
@@ -0,0 +1,12 @@
+/**
+ *
+ * @param {String} text
+ * @param {Object} settings
+ * @returns {string|*|XML|void}
+ */
```
youknowriad: Can we fix the remaining JS Docs?

packages/wordcount/src/index.js
```diff
@@ -0,0 +1,104 @@
...
+ * @param {String} type         The type of count. Accepts ;words', 'characters_excluding_spaces', or 'characters_including_spaces'.
+ * @param {Object} userSettings Custom settings object.
+ *
+ * @return {Number} The word or character count.
+ */
+
+export function count( text, type, userSettings ) {
+	const settings = loadSettings( type, userSettings );
+	if ( text ) {
+		let matchRegExp = settings[type + 'RegExp'];
```
adamsilverstein: @ryanwelcher this line and the next could  use a little whitespace

PR #11, opened 2017-07-16
Add a11y and dom-ready packages.

omarreiss:
Adds the `a11y-speak` module with all of its commit history from https://github.com/Yoast/a11y-speak to the WordPress packages repo.

Steps taken:

- Did a `git subtree add prefix=packages/a11y-speak https://github.com/Yoast/a11y-speak.git master` to import the git history.
- Removed all the obsolete files.
- Updated package.json.
- Refactored the module to use `const`, `let`, and es6 imports. 
- I extracted most of the logic into separate modules to make it unit testable. 
- Wrote unit tests.

One thing that I think we should consider is replace [`domReady`](https://github.com/WordPress/packages/blob/add/a11y-speak/packages/a11y-speak/src/domReady.js) with an external dependency -> https://github.com/ded/domready, which is the most popular npm module for this. If we do want to keep our own implementation, that seems like a thing that should be its own module and not part of `a11-speak`.


adamsilverstein:
Looks good to me, nice work! 

I especially like the way you refactored and broke out the functions and imported them and also added all the tests. 

I also see why you just want to import domReady although the domReady module seems to do more than we need, our own implementation is simpler. The [module code](https://github.com/ded/domready/blob/master/src/ready.js) has a timeout block, maybe to address older ie? - we shouldn't need this since we only [support ie > 10](http://caniuse.com/#search=domcontentloaded). If you want to make it its own module, that would be good, it is likely to be useful elsewhere.

packages/a11y-speak/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "@wordpress/a11y-speak",
+  "version": "0.0.1",
+  "description": "Easily announce dynamic interface updates to screen readers using ARIA live regions.",
+  "homepage": "https://github.com/WordPress/packages/packages/a11y-speak/README.md",
```
aduth: This will 404, needs the `blob/master` path fragment. Instead:

`https://github.com/WordPress/packages/tree/master/packages/a11y-speak`

packages/a11y-speak/src/domReady.js
```diff
@@ -0,0 +1,16 @@
+/**
+ * Specify a function to execute when the DOM is fully loaded.
+ *
+ * @param {Function} callback A function to execute after the DOM is ready.
+ *
+ * @returns {void}
+ */
+const domReady = function( callback ) {
+  if ( document.readyState === 'complete' || ( document.readyState !== 'loading' && !document.documentElement.doScroll ) ) {
```
aduth: There should be a space after the `!` exclamation:

>Any `!` negation operator should have a following space.

https://make.wordpress.org/core/handbook/best-practices/coding-standards/javascript/#spacing

We may need to update our ESLint configuration for `space-unary-ops` options:

http://eslint.org/docs/rules/space-unary-ops#options

packages/a11y-speak/README.md
```diff
@@ -0,0 +1,37 @@
...
+> That’s what `wp.a11y.speak()` is meant for. It’s a simple tool that creates and appends an ARIA live notifications area to the <body> element where developers can dispatch text messages. Assistive technologies will automatically announce any text change in this area. This ARIA live region has an ARIA role of “status” so it has an implicit aria-live value of polite and an implicit aria-atomic value of true.
+
+> This means assistive technologies will notify users of updates but generally do not interrupt the current task, and updates take low priority. If you’re creating an application with higher priority updates (such as a notification that their current session is about to expire, for example), then you’ll want to create your own method with an aria-live value of assertive.
+
+## Usage
+
+To make the `wp.a11y.speak` functionality more universally available, we've decided to create a dedicated JS module for it, called `A11ySpeak`. Usage is very simple:
+
+```JS
+var A11ySpeak = require('A11ySpeak');
```
aduth: This should be `require( '@wordpress/a11y-speak' );`.

Also thinking we should refer to this with lower-case assignment, since it's not a constructor:

>Variable and function names should be full words, using camel case with a lowercase first letter. _[...]_ Constructors intended for use with new should have a capital first letter (UpperCamelCase).

https://make.wordpress.org/core/handbook/best-practices/coding-standards/javascript/#naming-conventions

Instead: `var speak = require( '@wordpress/a11y-speak' )`

For that matter, do we really need to prefix `a11y-` ? It's hard to type out and `@wordpress/speak` would work just as well. Might need some thought as to how we align with the core global APIs `wp.a11y.speak`. Maybe the package should be `@wordpress/a11y`, of which `speak` is one of the named exports?

packages/a11y-speak/README.md
```diff
@@ -0,0 +1,37 @@
...
+// For polite messages that shouldn't interrupt what screen readers are currently announcing.
+A11ySpeak( 'The message you want to send to the ARIA live region' );
+
+// For assertive messages that should interrupt what screen readers are currently announcing.
+A11ySpeak( 'The message you want to send to the ARIA live region', 'assertive' );
+```
+
+## Browser support
+
+IE9+ and all modern browsers.
```
aduth: Do we have a good canonical link we could reference instead of inlining this text, in case this is ever updated in the future?

packages/a11y-speak/README.md
nylen: > Maybe the package should be `@wordpress/a11y`, of which `speak` is one of the named exports?

+1

packages/a11y-speak/package.json
omarreiss: Good one!

packages/a11y-speak/src/domReady.js
omarreiss: Will do!

packages/a11y-speak/README.md
omarreiss: Yes, was thinking the same thing. Will change it to `@wordpress/a11y` Will make speak a named export.

packages/a11y-speak/README.md
ntwb: This should do the trick https://make.wordpress.org/design/handbook/browser-support/

It's from where WP bases its _browserlist_ matrix, and updates as required

packages/a11y/src/index.js
```diff
@@ -0,0 +1,46 @@
+import addContainer from './addContainer';
+import clear from './clear';
+import domReady from '../../domReady/src/';
```
omarreiss: So this is quite horrible in a way. Until `domReady` is published I cannot 
```JS
import domReady from '@wordpress/domReady';
```

Unfortunately NPM also doesn't support installing an NPM module from a github monorepo subdir [and will not in the future](https://github.com/npm/npm/issues/2974). Don't know if you can think of anything @aduth? If not, I'll just make an issue to fix this after we publish this.

omarreiss: Not even sure if this will go well when publishing to NPM. Maybe that'd require me to first do separate PR for the domReady package.

youknowriad: I think there's some Lerna magic to leverage here. I think if we add a package to the dependencies of another package, it will automatically be symlinked if the versions match

omarreiss:
Now all I have to do is find a way to increase coverage for domReady package and then we're done.

packages/a11y/src/index.js
omarreiss: That. works. awesomely! Done, thanks!

packages/a11y-speak/README.md
omarreiss: Done!

omarreiss:
The way I've implemented it now, I need to 
```SH
lerna bootstrap && npm run build
```
before I can run the unit tests. This is because the `a11y` package needs to be able to import the `domReady` module. I expected this to resolve fine by simply mocking the module, but somehow this doesn't seem to work. 
Right now, the only way I know how to make Travis pass, is to add the above line to the `.travis.yml`. That feels far from ideal. 

@youknowriad @aduth can you think of anything I'm missing?

packages/domReady/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "@wordpress/domReady",
```
aduth: We'll need this to be dash-delimited:

>New packages must not have uppercase letters in the name.

https://docs.npmjs.com/files/package.json#name

aduth:
>Right now, the only way I know how to make Travis pass, is to add the above line to the .travis.yml. >That feels far from ideal.
>
>@youknowriad @aduth can you think of anything I'm missing?

From what I can tell, this would be necessary because npm alone isn't smart enough to delve into and install dependencies of bundled packages. Other Lerna projects do the same:

https://github.com/babel/babel/blob/b242e0d/.travis.yml#L18
https://github.com/babel/babel/blob/b242e0d/Makefile#L61
https://github.com/babel/babel/blob/b242e0d/Makefile#L83

packages/domReady/package.json
omarreiss: Whoops thanks, wasn't paying attention earlier.

packages/dom-ready/src/index.js
```diff
@@ -0,0 +1,16 @@
+/**
+ * Specify a function to execute when the DOM is fully loaded.
+ *
+ * @param {Function} callback A function to execute after the DOM is ready.
+ *
+ * @returns {void}
+ */
+const domReady = function( callback ) {
+  if ( document.readyState === 'complete' || ( document.readyState !== 'loading' && !document.documentElement.doScroll ) ) {
```
omarreiss: @afercia since we don't have to support ie, this could just be

```JS
if ( document.readyState === 'complete' ) {
```
Right?

omarreiss: At least that's what I take from https://developer.mozilla.org/nl/docs/Web/API/Document/readyState

afercia: @omarreiss yes, checked a couple resources and I think that part can go.

omarreiss:
@youknowriad think this is good for merge. Did my best to maximize coverage. If you want, I can try even harder. But it's a lot of weird DOM stuff. Find myself hacking around a lot to get all the branches covered.

aduth:
We lack ESLint so far (#7), but the styling here needs some work, chiefly tabs vs. spaces (tabs).

omarreiss:
@aduth Checked against current ESlint config in #7, fixed all errors.

packages/a11y/src/test/index.test.js
```diff
@@ -0,0 +1,102 @@
...
+
+import clear from '../clear';
+import domReady from '@wordpress/dom-ready';
+import filterMessage from '../filterMessage';
+
+describe( 'speak', () => {
+	let containerPolite = document.getElementById( 'a11y-speak-polite' );
+	let containerAssertive = document.getElementById( 'a11y-speak-assertive' );
+
+	beforeEach( () => {
```
aduth: Should the function spacing exception apply here?

https://make.wordpress.org/core/handbook/best-practices/coding-standards/javascript/#arrays-and-function-calls

omarreiss: I guess Arrow functions are functions too. I'll apply the exception.

omarreiss:
Revisions done, coverage 100%, merging.

PR #12, opened 2017-07-18
Add a sanitize package

adamsilverstein:
These could use some unit tests.


packages/sanitize/src/index.js
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Strip HTML tags and convert HTML entities.
+ *
+ * @param {string} text Text to strip tags and convert HTML entities.
+ *
+ * @return Sanitized text. False on failure.
+ */
+export function sanitizeText( text ) {
+	var _text = wp.utils.stripTags( text ),
```
aduth: How should we handle this dependency? Ideally it'd be referenced by Node module, but at the very least we need to ensure that the corresponding script handle is enqueued whenever the sanitize package is used.

packages/sanitize/src/index.js
```diff
@@ -0,0 +1,34 @@
...
+		textarea = document.createElement( 'textarea' );
```
aduth: Could we reuse a shared reference of a textarea element? Is there any overhead to creating a new element every time this is invoked?

packages/sanitize/src/index.js
```diff
@@ -0,0 +1,34 @@
...
+
+	try {
+		textarea.innerHTML = _text;
+		_text = wp.utils.stripTags( textarea.value );
```
aduth: A potential optimization to consider is bypassing this step if there are no encoded entities to convert.

See related: https://github.com/Automattic/wp-calypso/pull/9725

aduth (commented):
Could also do for a README.md

packages/sanitize/src/index.js
```diff
@@ -0,0 +1,34 @@
+/**
+ * Strip HTML tags.
+ *
+ * @param {string} text Text to have the HTML tags striped out of.
```
nylen: Typo: `striped` should be `stripped`

packages/sanitize/src/index.js
```diff
@@ -0,0 +1,34 @@
...
+}
+
+/**
+ * Strip HTML tags and convert HTML entities.
+ *
+ * @param {string} text Text to strip tags and convert HTML entities.
+ *
+ * @return Sanitized text. False on failure.
+ */
+export function sanitizeText( text ) {
```
nylen: This needs a better name and description.  `sanitizeText` - in what way, against what? `convert HTML entities` - into what?

If I read this correctly, I think the answers are "remove tags and encode HTML entities".  I'm having a hard time thinking of a good function name, though.  Maybe we should just create the `encodeHTMLEntities` first.

nylen:
When are these functions expected to be used?  In particular the combination of "strip tags and encode HTML entities" seems a bit strange to me.

packages/sanitize/src/index.js
adamsilverstein: We need the combined function to maintain existing functionality in PressThis. I'll try to improve the naming and descriptions here. The unit tests would also help make the usages clearer.

packages/sanitize/src/index.js
adamsilverstein: right, thanks.

packages/sanitize/src/index.js
adamsilverstein: Good suggestion, for now the code here exactly matches what we have in core.

packages/sanitize/src/index.js
adamsilverstein: I'm sure there is some overhead and I like that suggestion, would that possible get destroyed though in  a SPA context?

packages/sanitize/src/index.js
adamsilverstein: oh, yea - good catch. i'll switch this to referencing as a node module

packages/sanitize/src/index.js
aduth: Calypso uses a single shared reference without any trouble:

https://github.com/Automattic/wp-calypso/blob/2bdafc81fb2e2d7fd6027ded360b9641d57564a0/client/lib/formatting/decode-entities/browser.js#L1-L7

ntwb:
I've added a handful of basic tests for `stripTags` via https://github.com/WordPress/packages/pull/12/commits/7e360f0fa7f5982d9833afbf28238eecb6727639

adamsilverstein:
> When are these functions expected to be used? In particular the combination of "strip tags and encode HTML entities" seems a bit strange to me.
@nylen -
In WordPress core, the functions are used in wp-admin/js/press-this.js. `sanitizeText` is called in `getTitleText` to safely set the untrusted title text: `$( '#post_title' ).val( getTitleText() );`. `stripTags` is used in that function and also might be generally useful: it uses regexs to strip tags from content.

packages/sanitize/src/index.js
adamsilverstein: Addressed in https://github.com/WordPress/packages/pull/12/commits/4eb3d86dd59bd4bc89183908453eef98e9c6e374

packages/sanitize/src/index.js
adamsilverstein: this has been refactored and is now imported

packages/sanitize/src/index.js
adamsilverstein: I updated the inline docs a bit to hopefully clear up how these functions work.

packages/sanitize/src/index.js
adamsilverstein: Fixed in https://github.com/WordPress/packages/pull/12/commits/6c93f801d6fd0138491e434775f13382b1bfad68

notnownikki:
I just added a `decodeEntities` util to Gutenberg, and it was suggested it might be relevant here too.

Commit is at https://github.com/WordPress/gutenberg/commit/f08c5f46b690e7bd0b53aa8ac93169e3a3681ec9 , if it's useful I'm happy to work on contributing it here.

adamsilverstein:
@notnownikki that does seem useful, not sure it belongs with `sanitize`? I had originally put these helpers in wp.utils, only later separating them out.

ntwb:
Issue moved to [WordPress/gutenberg #7824](https://github.com/WordPress/gutenberg/issues/7824) via [**ZenHub**](https://www.zenhub.com/)

PR #13, opened 2017-07-18
Add the WordPress hook library from 21170-core

adamsilverstein:
See https://core.trac.wordpress.org/ticket/21170


packages/hooks/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "@wordpress/hooks",
+  "version": "1.0.0",
```
nylen: Let's make this `0.1.0` or something, I don't think we're at "final" v1 yet.

packages/hooks/src/test/index.test.js
```diff
@@ -0,0 +1,4 @@
+/**
+ * Internal Dependencies
+ */
+import { doAction, applyFilters, addAction, addFilter, doingAction, doingFilter, didAction, didFilter, hasAction, hasFilter, removeAllActions, removeAllFilters, currentFilter } from '../';
```
nylen: Let's split this long line up into one import per line.

packages/hooks/src/index.js
```diff
@@ -0,0 +1,333 @@
...
+export doAction =     createRunHookByType( 'actions', runDoAction );
+export applyFilters = createRunHookByType( 'filters', runApplyFilters );
+
+// Add functions.
+export addAction = createAddHookByType( 'actions' );
+export addFilter = createAddHookByType( 'filters' );
+
+// Doing functions.
+export doingAction = createDoingHookByType( 'actions' ), /* True for actions until;next action fired. */
+export doingFilter = createDoingHookByType( 'filters' ), /* True for filters while;filter is being applied. */
```
nylen: Why the semicolons in the middle of these comments?  Should be `//` inline comments as well

packages/hooks/src/index.js
```diff
@@ -0,0 +1,333 @@
...
+			false;
+	};
+}
+
+/**
+ * Remove all the actions registered to a hook,
+ */
+function createRemoveAllByType( type ) {
+	return function( action, type ) {
+
```
nylen: No implementation yet?

packages/hooks/src/test/index.test.js
rmccue: We could also `export default { doAction, ... }` and then just `import hooks from '../`, would likely be a bit nicer.

nylen: The drawback there is that you have to prefix every function call with `hooks.`

packages/hooks/package.json
adamsilverstein: Ok!

packages/hooks/src/index.js
adamsilverstein: fixed on the ticket, will update...

packages/hooks/src/index.js
adamsilverstein: thanks for noting, will correct

packages/hooks/src/test/index.test.js
adamsilverstein: I'll work on cleaning this up and maybe splitting out the generator code into files/export/import.

adamsilverstein:
>The machinery around createSomeFunction( type ) and all the corresponding HOOKS[ type ] checks feels a bit overwrought (and will perform more slowly than necessary). 

@nylen I'm not sure I understand how exactly you are proposing refactoring this and how it would help... the `type` part is a wrapper around the generators to account for our two hook types.

nylen:
> @nylen I'm not sure I understand how exactly you are proposing refactoring this and how it would help... the `type` part is a wrapper around the generators to account for our two hook types.

Instead of passing around `type` strings, pass around the actual objects that need to be manipulated instead (`HOOKS.actions` and `HOOKS.filters` in the current implementation).

This should make the code clearer, shorter, and faster.

adamsilverstein:
>Instead of passing around type strings, pass around the actual objects that need to be manipulated instead (HOOKS.actions and HOOKS.filters in the current implementation).

Ah, ok - interesting idea, I can see how that would be cleaner and faster. I'll give it a try on the core patch where I have unit tests that will tell me if I broke anything.

adamsilverstein:
@nylen something like this? https://core.trac.wordpress.org/attachment/ticket/21170/21170.11.diff

nylen:
Yes, except several operations are `O(total number of hooks)` and this should be `O(total number of hooks with the given name)` instead. Let's move development including unit tests to here - I can help with that tomorrow.

adamsilverstein:
@nylen 
> several operations are O(total number of hooks) and this should be O(total number of hooks with the given name) instead. 

Not sure what you mean by this, can you clarify? 

I reworked the tests as Jest tests.

Thinking of breaking out the code into files and maybe temporarily adding a build step to generate the core file so I don't have to keep updating both.

adamsilverstein:
I broke out the code a bit and tested the built in build process. What is the best way to get a build step that sets up a single file we can load into core that loads functions on wp.hooks? Do we need webpack with an entry file like this? https://gist.github.com/adamsilverstein/4fbbd4ca14c5f7526a1ee41b63509cd2

nylen:
Webpack can do that part for us, like the Gutenberg build does:  https://github.com/WordPress/gutenberg/blob/c78ad83/webpack.config.js#L49-L74

That should probably be handled at the root level of this repo, though also related is how we'll publish the package on `npm`.

adamsilverstein:
Ok, i'll work on the build part in the root with webpack, can we publish the package as @wordpress/hooks - and do we let people import individual functions, or only the entire hooks object?

nylen:
People should be able to import individual functions.

adamsilverstein:
Ok, that is how we have it. I'll work on the wp.hooks object for core as part of the build.

adamsilverstein:
@nylen changes look good so far, thanks!

nylen:
> > several operations are O(total number of hooks) and this should be O(total number of hooks with the given name) instead. 

> Not sure what you mean by this, can you clarify? 

I was misreading the code here, the declaration of `HOOKS.actions` and `HOOKS.filters` as arrays `[]` rather than objects `{}` was throwing me off.  Fixed that in bee29d7.

nylen:
Regarding the build process, it looks like the existing `npm run build` will not end up serving our needs so we should replace it with Webpack in a separate PR.

adamsilverstein:
> I was misreading the code here, the declaration of HOOKS.actions and HOOKS.filters as arrays []rather than objects {} was throwing me off. Fixed that in bee29d7.

I wasn't sure these would work as objects rather than arrays because we use things like `push` and `splice`. Everything still works after the change though, so I think we are fine. Maybe the arrays are one level deeper or these calls work with objects?

nylen:
The arrays are one level deeper.

nylen:
I am generally pretty happy with this (the next thing I want to do is remove the unnecessary and slow `sortHooks` function).

A couple of high-level questions, though.

## Different types of hooks

Do we really need both actions and filters?  A filter is a superset of an action.  If we were re-designing this system today, would we include actions at all?

If we can just use filters for everything, the code will become a lot simpler.

### `_doing_it_wrong` and i18n

In e3392c3 I added error reporting for failure cases.  In general we should be really thorough and strict about validating parameters passed to these functions, and include test cases for these failure branches, like we do in Gutenberg for the block registration APIs.

These `console.error` statements look a lot like `_doing_it_wrong`, and they need to be translated.  This makes me think we need to include `i18n` and something like `wp._doingItWrong()` in this repo sooner rather than later.

adamsilverstein:
> Do we really need both actions and filters? 

The distinction seems useful even though internally they function the same way... especially coming from the existing PHP hooks system.  The naming semantics makes it clear what the purpose of the hook is.

Action hooks are for anything that developers would want to respond to - an event occurring for example adding a new menu in the wp-admin menu editor  - probably anywhere we currently use `$( document ).trigger( ...` in core today. Actions do not respond to returned values.

Filter hooks would be for any default object developers should be able to filter - for example a default or returned value. Filters always expect a returned value.

adamsilverstein:
I added a webpack build and verified bringing this file back into core works as expected. my qunit tests for core needed a little adjusting for some changes we made here, just as not removing all hooks unless explicitly calling removeAll (which I'm still not certain about).

packages/hooks/src/createAddHook.js
```diff
@@ -0,0 +1,64 @@
...
+			return;
+		}
+
+		if ( typeof callback !== 'function' ) {
+			console.error( 'The hook callback must be a function.' );
+			return;
+		}
+
+		// Assign default priority
+		if ( 'undefined' === typeof priority ) {
```
aduth: Minor: Could be surfaced as default in arguments:

```js
return function addHook( hookName, callback, priority = 10 ) {
```

packages/hooks/src/createRemoveHook.js
```diff
@@ -0,0 +1,44 @@
...
+	 *                             all callbacks.
+	 */
+	return function removeHook( hookName, callback ) {
+		// Bail if no hooks exist by this name
+		if ( ! hooks.hasOwnProperty( hookName ) ) {
+			return;
+		}
+
+		if ( removeAll ) {
+			const runs = hooks[ hookName ].runs;
```
aduth: Minor: Could be destructured:

```js
const { runs } = hooks[ hookName ];
```

packages/hooks/src/createRemoveHook.js
```diff
@@ -0,0 +1,44 @@
...
+			hooks[ hookName ] = [];
+			if ( runs ) {
+				hooks[ hookName ].runs = runs;
```
aduth: Makes me slightly uncomfortable that we assign this as a property of an array, versus separating `hooks[ hookName ].handlers = []` and `hooks[ hookName ].runs = 0;`. Seems like this could simplify some of the juggling to preserve runs here too. Current logic appears to be fine, but future revisions which iterate by `for ( const key in` could yield unexpected keys.

packages/hooks/src/createRunHook.js
```diff
@@ -0,0 +1,47 @@
...
+	 * @param  {string} hookName The name of the hook to run.
+	 * @param  {...*}   args     Arguments to pass to the hook callbacks.
+	 *
+	 * @return {*}               Return value of runner, if applicable.
+	 */
+	return function runHooks( hookName, ...args ) {
+		const handlers = hooks[ hookName ];
+		let maybeReturnValue = args[ 0 ];
+
+		if ( ! handlers ) {
```
aduth: Along lines of previous comment about separating `handlers`, since `removeAll` flow in `removeHook` will reset handlers to an empty array, this condition won't trigger. Would we expect `runs` to increment after all handlers are removed for a hook? Otherwise this would at least need to be changed to `if ( ! handlers || ! handlers.length )` (since empty array is truthy).

packages/hooks/package.json
```diff
@@ -0,0 +1,18 @@
...
+    "type": "git",
+    "url": "https://github.com/WordPress/packages.git"
+  },
+  "description": "WordPress Hooks library",
+  "main": "build/index.js",
+  "module": "build-module/index.js",
+  "browser": "build-browser/index.js",
+  "author": "WordPress",
+  "license": "GPL-2.0+",
+  "dependencies": {
```
aduth: Are we using either of these dependencies?

packages/hooks/package.json
```diff
@@ -0,0 +1,18 @@
+{
```
aduth: I'd suggest a `files` values to reduce bloat of distributed module:

https://docs.npmjs.com/files/package.json#files

e.g. omitting tests.

webpack.config.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
```
aduth: - Should this be integrated into the existing build step?
- Since we create many different build variants anyways, is `hooks/build/wp-core` the best location, or should we have a `hooks/build-core` or similar?
- Instead of explicitly setting `window` globals in `src/wp-hooks.js`, you might consider [Webpack's `library` and `libraryTarget` options](https://webpack.js.org/configuration/output/#output-library)
- The gap has narrowed in recent Webpack releases, but I've tended toward Rollup as a bundler for library code specifically ([related article](https://medium.com/webpack/webpack-and-rollup-the-same-but-different-a41ad427058c)). Not really pushing it, but raising for discussion nonetheless.

aduth:
Oh, can we add a README.md as well (for any and all modules).

webpack.config.js
nylen: -1 on having a separate webpack config file for each module, this should be handled at the repo level (and in a separate PR).

Why do we have so many different build variants?

nylen:
> my qunit tests for core needed a little adjusting for some changes we made here

Why do we need the qunit tests if we have tests here?

> just as not removing all hooks unless explicitly calling removeAll (which I'm still not certain about).

I feel quite strongly that `removeAll` should be an explicitly requested action, rather than an additional behavior of `removeFilter`:

- PHP filters [do not support](https://codex.wordpress.org/Function_Reference/remove_filter#Parameters) this additional behavior
- It's duplicating existing functionality - there should be only one way to do this
- This is going to lead to bugs when developers accidentally pass a falsy value to `removeFilter`.

nylen:
In general the nested `package.json` thing is pretty annoying, it means that `npm test` and any other scripts don't work unless you change back to the parent directory.  Curious what we can do to fix that, short of duplicating all the npm scripts in each subdirectory.

packages/hooks/src/createRunHook.js
nylen: Currently this will trigger when a hook run is attempted but no callbacks have ever been added for that hook.

However, `runs` should increment if no handlers have ever been set, or if they have all been cleared, good catch there.  This will match PHP behavior.

packages/hooks/src/createRemoveHook.js
nylen: I think this approach will be cleaner as well.  However, I think we have to avoid `for ... in` because we can't guarantee that plugins aren't extending object prototypes.

packages/hooks/src/createRunHook.js
```diff
@@ -0,0 +1,56 @@
...
+
+		let maybeReturnValue = args[ 0 ];
+		handlers.forEach( handler => {
+			maybeReturnValue = handler.callback.apply( null, args );
+			if ( returnFirstArg ) {
+				args[ 0 ] = maybeReturnValue;
+			}
+		} );
+
+		delete hooks.current;
```
nylen: This will mean that nested filter calls fail to return the correct value for `currentFilter()`.  This needs to be a stack instead.

nylen:
There is a surprising amount still left to do here before this is ready to ship.  Maybe doesn't all have to be done in this PR, but definitely before release.  Here's what I can think of and what's left from the review on this PR so far:

- [ ] I agree we need tests for removing a hook during its run when the removed hook's priority is before the current callback
- [ ] same but when the removed callback is the same as the current callback
- [ ] same but when the removed callback would come after the current callback (then I believe it should not be called, but let's double check with the PHP implementation)
- [ ] I agree we need tests for running hooks recursively; let's look at the PHP code to see what special handling this case needs
- [ ] change `hooks.current` into a stack as commented above
- [ ] consider changing `hooks.current` to `hooks.__current` and disallowing hook names beginning with `__`; otherwise having a hook named `current` would cause unintended results
- [ ] look through the PHP implementation and unit tests (and talk to people who've worked on it) about what other special cases we may need to handle
- [ ] centralize the webpack build to the entire repo
- [ ] add a readme
- [ ] add a `files` section to `package.json`

adamsilverstein:
+1 to your suggestion around making current a stack and renaming it __current

ntwb:
See the i18n [comment above](https://github.com/WordPress/packages/pull/13#issuecomment-316770832) @joyously

adamsilverstein:
I started on some of the todos:

* added some tests around removing hooks while a callback is running, 
* found a case where the filters didn’t work as expected, I left in a failing test for reference
* added a readme
* changed `current` to `__current` and made it a stack - that could use some review & testing
* we could use recursive testing for doingAction/Filter, currentAction/Filter

my main goal with webpack so far is to have a way - for now - to build a release file for WordPress core. I want to get wp.hooks into core trunk sooner rather than later so developers and the core team can begin to try it out.

Ultimately I may move this build functionality to core directly - once we have webpack in place and hooks in npm we can import and build directly as part of our core build process.

joyously:
> Ultimately I may move this build functionality to core directly - once we have webpack in place and hooks in npm we can import and build directly as part of our core build process.

Would this mean that the WordPress code base would be located in multiple places? Isn't this a new thing? Is it a wanted thing?
Or would SVN have the definitive version and it gets copied to npm?

packages/hooks/src/index.js
```diff
@@ -27,8 +27,8 @@ export const doAction     = createRunHook( HOOKS.actions );
 export const applyFilters = createRunHook( HOOKS.filters, true );
 
 // Current action/filter functions.
-export const currentAction = () => HOOKS.actions.current || null;
-export const currentFilter = () => HOOKS.filters.current || null;
+export const currentAction = () => HOOKS.actions.__current[0] || null;
+export const currentFilter = () => HOOKS.filters.__current[0] || null;
```
nylen: This should return the last element instead of the first, and it needs unit tests for multiple filters running at the same time.  It's probably worth splitting it out into a separate file again.

nylen:
Let's disallow any hook name starting with `__` and print a `console.error` message.  We may want to attach extra properties there in the future.

nylen:
Let's also remove the global functions in the tests and make them inline `lambda => {}` functions.  They're not very long, and each test should be readable on its own.

packages/hooks/src/test/index.test.js
```diff
@@ -239,6 +239,15 @@ test( 'remove a filter callback of higher priority when running hook', function(
...
 	expect( applyFilters( 'test.filter', 'test' ) ).toBe( 'testabc' );
 } );
+test( 'remove one hook in a callback before adding another', function() {
+	addFilter( 'test.filter', filter_a, 1 );
+	addFilter( 'test.filter', filter_b, 3 );
+	addFilter( 'test.filter', filter_c, 5 );
+	addFilter( 'test.filter', filter_a_removesb, 4 );
+
+	// Fails!!! only a & b run, not c.
+	expect( applyFilters( 'test.filter', 'test' ) ).toBe( 'testabc' );
```
nylen: In the latest build, this appears to be passing?

nylen: Ah I see, the correct assertion is `testab`.

packages/hooks/src/test/index.test.js
```diff
@@ -239,6 +239,15 @@ test( 'remove a filter callback of higher priority when running hook', function(
 
 	expect( applyFilters( 'test.filter', 'test' ) ).toBe( 'testabc' );
 } );
```
nylen: Minor: Let's put a blank line in between each test.

nylen:
> Would this mean that the WordPress code base would be located in multiple places? Isn't this a new thing?

Yes, and yes.

> Is it a wanted thing?

Please :smile: It's been roughly 1,000,000x easier to iterate on this piece of code here on GitHub than it would have been on Trac.

nylen:
> In part because I want to make sure the built file we bring over to core works as expected, so far the initial webpack build passed tests, but not eslint

I don't think it's necessary for the built file to pass eslint.  It's generated code and shouldn't be modified directly.

> I want to get wp.hooks into core trunk sooner rather than later so developers and the core team can begin to try it out.

There is **a lot** that still needs to be done before this can ship in core.  For example, recursion handling ([start looking here](https://core.trac.wordpress.org/browser/tags/4.8/src/wp-includes/class-wp-hook.php?marks=340,341,344,348,350,351#L331)), modifying the list of hooks during a callback run ([start looking here](https://core.trac.wordpress.org/browser/tags/4.8/src/wp-includes/class-wp-hook.php#L98)), and [the related unit tests](https://core.trac.wordpress.org/browser/tags/4.8/tests/phpunit/tests/hooks).

I think it would be better to publish an initial version of this code to `npm` (like `0.0.1-alpha`) and start using it in Gutenberg.  That way we don't create any kind of expectation that this is ready for production use.

> Let's also remove the global functions in the tests and make them inline `lambda => {}` functions.  They're not very long, and each test should be readable on its own.

Looking closer, this probably won't make sense to do everywhere, because some of these callbacks are heavily re-used and accessed by reference equality to remove them from the callback list.

nylen:
Recursion in filters should be allowed rather than prohibited.  It's up to the filter callback author to implement it correctly and avoid infinite loops.

Apply this patch to core to try out the PHP behavior:

```diff
diff --git a/tests/phpunit/tests/hooks/add_filter.php b/tests/phpunit/tests/hooks/add_filter.php
index 79aeb75..10c4be0 100755
--- a/tests/phpunit/tests/hooks/add_filter.php
+++ b/tests/phpunit/tests/hooks/add_filter.php
@@ -155,6 +155,27 @@ class Tests_WP_Hook_Add_Filter extends WP_UnitTestCase {
 		$this->assertSame( '12', $value );
 	}
 
+	public function recurse_unlimited( $value ) {
+		return apply_filters( 'test_recurse_unlimited', $value . 'X' );
+	}
+
+	public function test_recurse_unlimited() {
+		add_filter( 'test_recurse_unlimited', array( $this, 'recurse_unlimited' ) );
+		$this->assertEquals( 'testX', apply_filters( 'test_recurse_unlimited', 'test' ) );
+	}
+
+	public function recurse_limited( $value ) {
+		if ( strlen( $value ) === 7 ) {
+			return $value;
+		}
+		return apply_filters( 'test_recurse_limited', $value . 'X' );
+	}
+
+	public function test_recurse_limited() {
+		add_filter( 'test_recurse_limited', array( $this, 'recurse_limited' ) );
+		$this->assertEquals( 'testXXX', apply_filters( 'test_recurse_limited', 'test' ) );
+	}
+
 	public function test_remove_and_recurse_and_add() {
 		$this->hook = new Wp_Hook();
 
```

Then:

- `phpunit --filter test_recurse_unlimited`: runs out of memory
- `phpunit --filter test_recurse_limited`: passes (added a similar, failing test to this PR in bd911d0)

The tests immediately below the ones in this core patch - starting at [`test_remove_and_recurse_and_add`](https://core.trac.wordpress.org/browser/tags/4.8/tests/phpunit/tests/hooks/add_filter.php#L158) - are the tricky recursion-related behavior that we'll need to support.  Getting similar, failing tests added here is a good next step.

youknowriad:
I'm missing a lot of context regarding this module so please forgive me if this is already answered somewhere.

Unlike PHP, I suspect we need to handle Async hooks somehow, What If a filter needs to trigger a network call before returning its value? Is this already handled?

nylen:
> Unlike PHP, I suspect we need to handle Async hooks somehow, What If a filter needs to trigger a network call before returning its value? Is this already handled?

Yes, it's a good question... and no, this is not currently handled.

We could probably use promises in some interesting ways to help solve this problem.

nylen:
Initial investigation into core behavior handling adding/removing filter callbacks while a filter is running reveals that it is... much weirder than I expected.  Here is some light bedtime reading for interested parties:  https://core.trac.wordpress.org/ticket/17817

I don't understand why the test added here in 6653227 should pass.  To me it seems that it should be an infinite loop:

1. Execute `val => val + '1'` (callback 1 of 4).
2. Execute `removeRecurseAndAdd2` (callback 2 of 4).
   1. Remove `removeRecurseAndAdd2` (we've executed 1 of 3 callbacks).
   2. Execute the filter recursively; yields `134`
   3. Re-add `removeRecurseAndAdd2` (we've executed 1 of 4 callbacks).
3. Execute the next callback (2 of 4; go back to step 2).
4. If you get here, you can go relax on the beach for the rest of your life. :tropical_drink: 

In 2e9b96e I've implemented what I think this behavior **should** be (needs more tests) and in 54f70c1 I've hacked on the problematic test until it passes.

packages/hooks/src/wp-hooks.js
```diff
@@ -1,41 +0,0 @@
-import HOOKS from './hooks';
```
adamsilverstein: The point of this entry file is to build the hooks library for inclusion in core; maybe we can do this directly in core once we add webpack to our build chain. Can you explain how you see this logic being generated with Webpack?

nylen:
> The point of this entry file is to build the hooks library for inclusion in core; maybe we can do this directly in core once we add webpack to our build chain. Can you explain how you see this logic being generated with Webpack?

See https://github.com/WordPress/packages/pull/13#issuecomment-316701215.  This Webpack configuration generates code like the following:

```js
this["wp"] = this["wp"] || {}; this["wp"]["utils"] =
/******/ (function(modules) { // webpackBootstrap
// ... (more auto-generated code which attaches exported functions to the `utils` object)
```

We may need to tweak the way that logic is generated so that it doesn't overwrite existing objects (`wp.utils` in that example; see also https://github.com/WordPress/gutenberg/pull/1624#issuecomment-312420877).  However, I think it's pretty important that we avoid manually generating and maintaining files like that.  Handling imports/exports automatically seems like a key part of proving that we can use Webpack and ES6 modules in a WP context.

I think the build process should probably live in this repository, at least at first.

adamsilverstein:
@nylen Thanks for the webpack feedback, I will try to work on the webpack approach you suggested, as well as starting to think about how we import and build these packages into wp directly as part of our build process there.

packages/hooks/README.md
```diff
@@ -0,0 +1,34 @@
...
+* `removeAllActions(  'namespace.identifier' )`
+* `removeAllFilters(  'namespace.identifier' )`
+* `doAction( 'namespace.identifier', arg1, arg2, moreArgs, finalArg )`
+* `applyFilters( 'namespace.identifier', content )`
+* `doingAction( 'namespace.identifier' )`
+* `doingFilter( 'namespace.identifier' )`
+* `didAction( 'namespace.identifier' )`
+* `didFilter( 'namespace.identifier' )`
+* `hasAction( 'namespace.identifier' )`
+* `hasFilter( 'namespace.identifier' )`
```
kadamwhite: As discussed during core JS office hours today, I'd like to see us alter how we refer to these components -- in jQuery and other libraries "namespace" refers to the your-code-specific part following the period, and the event name is the part before the period. I believe that breaking from that convention would lead to a lot of confusion.

What I would propose instead:
```js
addAction( 'actionname.namespace' );
```

nylen: Yes, action name should come first if that is the way we want to go.  However, if it's required, maybe we should just make it a separate argument.  See discussion starting at https://core.trac.wordpress.org/ticket/21170#comment:129.

nylen:
## Event namespaces

From @aduth [on Slack](https://wordpress.slack.com/archives/C5UNMSU4R/p1501680389032946):

> I’d like to see some consistency with Gutenberg blocks if they’re both achieving a similar namespacing purpose. Something to keep in mind at least.

I think this is a good idea - these namespaces are intended to serve almost exactly the same purpose.  A Gutenberg block type name is `my-plugin-slug/block-name` ([ref](https://github.com/WordPress/gutenberg/blob/v0.6.0/blocks/api/registration.js#L29-L51)); a hook namespace could then be `my-plugin-slug/functionDescription`.

That implies the following function signature:

```js
/* function addFilter( filterName, namespace, callback ) { ... } */

addFilter( 'filter_name', 'my-plugin-slug/functionDescription', value => {
    // ...
    return value;
} );
```

In WP core, the namespace would always be `core/something`.

From the perspectives of both usage and implementation, this is much simpler than jQuery namespaces.  Those also have some [wrinkles](http://api.jquery.com/on/) that we don't really need:

> Namespaces are similar to CSS classes in that they are not hierarchical; only one name needs to match.

As I also noted [during the same Slack discussion](https://wordpress.slack.com/archives/C5UNMSU4R/p1501610251419518), it looks like jQuery-style namespaces were [initially created](https://github.com/jquery/jquery/commit/456f0fe5985eaad42eba456b9ff710771669607c) to solve the requirement of removing already-added callbacks in an existing API.  Let's add a clear and obvious parameter for this purpose from the beginning.  We can always relax this restriction later if needed; we cannot add new restrictions once people are already using this API.

nylen:
## Async support

I think if this is needed, it should work by returning a promise from a filter.

The details are a bit unclear to me and we need to work through some use cases first.  However, for now, we can add a `_doing_it_wrong`-style check and abort a filter run if a promise-like object is returned from a callback.  This will allow us to move forward with the changes in this PR and reserve this feature for future implementation.

Async support may not even be needed at all.  For the example of a network request, we could filter **both** the request parameters (including whether the request is sent) **and** the result of the request using two separate hooks, both synchronous.  We would want to create a standardized pattern for this.

kadamwhite:
> Let's add a clear and obvious parameter for this purpose from the beginning.

Sounds reasonable to me, I'm convinced 

> For the example of a network request, we could filter both the request parameters (including whether the request is sent) and the result of the request using two separate hooks

This sounds much more manageable to me than async actions. I don't believe I've heard a compelling argument for why returning a promise would be needed for which two actions wouldn't solve the problem just as well.

I wonder if there would be an easy way for us to use those before-request hooks to hijack and stop the outbound request and make another request or take some other action, such as pulling from a cache?

nylen:
> I wonder if there would be an easy way for us to use those before-request hooks to hijack and stop the outbound request and make another request or take some other action, such as pulling from a cache?

In PHP there is not really an accepted pattern for this either, but one that makes sense is to use a filter with a default value of `null` or `false`.  If some other value is returned by a filter callback, then the request is assumed to have been completed some other way.

This won't work in JavaScript because the request is asynchronous.  I see a few possible solutions:

1. Use the same `addFilter` API, but make execution asynchronous if a promise is returned from a filter.  We quickly enter the nightmare zone if some callbacks return promises and some do not.
2. Use `addFilterAsync` which passes a `next` callback to each filter callback rather than expecting a return value.  We quickly enter the nightmare zone when callbacks throw exceptions or forget to call their `next` function.
3. Use `addFilterAsync` which requires that all filter return values be promises.  <s>We quickly enter the nightmare zone</s> I am sure that this is not as simple as it sounds - we need to work through a use case.
4. Perform all needed functions with synchronous filters.  This will most likely involve multiple hooks.  We quickly enter the nightmare zone unless we have a strong convention or a bit of API to help name these hooks and manage how they relate to each other.

For the last option, and the example of a network request, there would be two hooks.  The first would accept an object of request arguments, be able to modify them by reference, and return `true` to continue normal execution or `false` to indicate that it will handle the request itself.  If a hook returns `false` then it must be responsible for calling some other function after the request completes.  This other function could be another hook, or it could be a `success` or `error` function passed as part of the request arguments.

The second hook (if present) would handle the response from the network request.  It would be helpful to have a common convention or a bit of API here to name the multiple hooks/callbacks involved and manage how they relate to each other.

Either way, this is something we'll need to work through with an example.  I don't see much here that should affect this PR, unless we want to reserve some features to prepare for a future implementation using the same API:

- Prohibit promises as return values of filters (option 1)
- Prohibit filter names beginning with `async_` (options 4 and/or 1)

packages/hooks/src/createAddHook.js
```diff
@@ -0,0 +1,71 @@
...
+	 * @param {Function} callback Function to call when the hook is run
+	 * @param {?number}  priority Priority of this hook (default=10)
+	 */
+	return function addHook( hookName, callback, priority = 10 ) {
+		if ( typeof hookName !== 'string' ) {
+			console.error( 'The hook name must be a string.' );
+			return;
+		}
+
+		if ( /^__/.test( hookName ) ) {
```
nylen: In the interest of starting with a restrictive API that allows for expansion later, let's restrict hook names to something like `/^[a-z][a-z0-9_]/`.

packages/hooks/src/createRunHook.js
```diff
@@ -0,0 +1,80 @@
...
+	/**
+	 * Runs all callbacks for the specified hook.
+	 *
+	 * @param  {string} hookName The name of the hook to run.
+	 * @param  {...*}   args     Arguments to pass to the hook callbacks.
+	 *
+	 * @return {*}               Return value of runner, if applicable.
+	 */
+	return function runHooks( hookName, ...args ) {
+		if ( 'string' !== typeof hookName ) {
```
nylen: The [JavaScript coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/javascript/#equality) don't recommend Yoda conditions.  In fact, the documentation and examples are mostly written in the opposite style.

Requiring `===` and `!==` removes the need for yoda conditions anyway, so I would like it if we could consider that requirement as entirely outdated anyway.

aduth:
>The [JavaScript coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/javascript/) don't recommend Yoda conditions. In fact, the documentation and examples are mostly written in the opposite style.
>
>Requiring `===` and `!==` removes the need for yoda conditions anyway, so I would like it if we could consider that requirement as entirely outdated anyway.

Not that I feel particularly strongly about them, but the guidelines do encourage Yoda (in specific circumstances? Under which the referenced line applies)

https://make.wordpress.org/core/handbook/best-practices/coding-standards/javascript/#yoda-conditions

nylen:
> the guidelines do encourage Yoda (in specific circumstances? Under which the referenced line applies)

I guess I missed this section of the guidelines via wishful thinking.  Thanks for pointing out my mistake.

Let's be consistent either way, then - we should either update all similar lines or none of them.

youknowriad:
Can I ask, why we're adding a webpack build to this PR? Can't this package be consumed like the other packages?

Not saying we shouldn't add a build script generating the globals (Though maybe it should be part of the core webpack config instead), but if we do something like that, it should be handled and published in the exact same way as the current builds (using the `build` script). And anyway, I think this should be added in a separate PR if we were to add it.

adamsilverstein:
@youknowriad The build process is here temporarily so I can build a wp.hooks file for core, I can remove it (and move to a separate PR if we want it available) before we merge the PR. Ultimately,  I plan to move the build/bundle to wp-core itself once we switch to webpack as part of our build process and the library is in npm.

adamsilverstein:
@kadamwhite I updated the Readme to reflect the new language and signature in https://github.com/WordPress/packages/pull/13/commits/4ece207a8465dce14ca5f2cab9d3bcc9f0c5a06f

adamsilverstein:
In the core JS chat today @schlessera suggested we consider adding a vendor prefix as part of the namespace string: https://wordpress.slack.com/archives/C5UNMSU4R/p1502803194000254 - this eems like a worthwhile addition that could prevent naming conflicts down the line.

schlessera:
As discussed in Slack, I'd like to suggest avoiding potential conflicts by changing the namespace prefix from `<package>/<function>` to `<vendor>/<package>/<something>`.

The reasoning behind this is that, for similar solutions, people tend to pick similar names. This makes the most important namespace level (`<package>` in this case) very prone to conflicts.

Adding the `vendor` prefix in front of that is a best practice in use in many different languages/frameworks, amongst others in the [larger PHP ecosystem](https://getcomposer.org/doc/01-basic-usage.md#package-names), the [.NET ecosystem](https://msdn.microsoft.com/en-us/library/893ke618(v=vs.71).aspx) or even GitHub repositories like this one: `WordPress\packages`.

For the example of the WP Core hooks above, you'd have `wordpress\core\something` instead. This opens the way for `wordpress\p2\something` or any other official package that is part of the WordPress vendor namespace.

I do understand that you want to keep the strings as short as possible, but maybe it would make sense to provide an automated/more convenient mechanism for _creating_ unique hook name instead in this case.

You could then have 1 central location where one sets up the namespace to use. People that need more control can still do it the more traditional way.

adamsilverstein:
@schlessera - I updated the namespace as per your suggestion, thanks!

packages/hooks/README.md
adamsilverstein: @kadamwhite this has been updated, can you plese review the current format?

adamsilverstein:
@nylen @kadamwhite @youknowriad @aduth - I'm planning to merge this to WordPress core soon so it has some time to soak for 4.9. Any final feedback here? 

Can we go ahead and merge this and release as a package version 1.0.0?

aduth:
I worry this was driven by a desire to mimic PHP, "what if _x_"s and "maybe it'd be nice if _x_" than it was by real-world requirements, ease-of-use, or expectations via prior art in JavaScript, culminating in awkward semantics around vendoring, namespaces, and hook names. If you count each of these as a pseudo-argument of sorts, it requires a minimum of 5 arguments to hook anything. I think hook naming was a valid direction to explore, but not one that ultimately had a positive outcome.

In place of naming to remove hooks, I'd rather we had taken the direction of some precedent:

- As in DOM and Node's `removeListener` and `removeEventListener`, passing a reference to the original callback
   - [`EventTarget.removeEventListener`](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener)
   - [`require( 'events' ).removeListener`](https://nodejs.org/api/events.html#events_event_removelistener)
- As in `clearTimeout`, `clearInterval`, and `cancelAnimationFrame`, passing the callback handle returned from the original hook
   - [`clearTimeout`](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/clearTimeout)
   - [`clearInterval`](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/clearInterval)
   - [`window.cancelAnimationFrame`](https://developer.mozilla.org/en-US/docs/Web/API/Window/cancelAnimationFrame)
- As in Redux subscribe, returning an unhook function as the return value of the original hook
   - [`Redux.subscribe`](http://redux.js.org/docs/api/Store.html#subscribe)

Or even a combination of the first and second.

Where are the requirements that dictated we should have addressable names for hooks? Even if we wanted to allow plugins to remove other plugin or core hook handlers, could it not be up to the plugins or core to create patterns around doing so, without needlessly bloating the argument signature of the hook creators?

Alternatively, even if we were to keep the direction we'd explored here, we might have considered some helpers to help create scoping to avoid repetition, like hook creators.

```js
const pluginHookCreator = createHookCreator( 'vendor/myplugin' );
pluginHookCreator.addAction( 'actionName', callback );
```

Or perhaps something integrated into build tooling that eliminates the need to explicitly specify the vendor, namespace. I'd like to explore similar for [JavaScript i18n string extraction](https://github.com/WordPress/gutenberg/tree/master/i18n).

In its current state, I'm not a fan of the approach here. Maybe we can discuss again in tomorrow's JS meeting.

pento:
This needs more than just time to soak in trunk before 4.9, it's a huge new API.

Please consider this a blocker: before it's merged, it needs to be integrated into a substantial project to help iron out practical problems that are likely to come up.

rmccue:
> In place of naming to remove hooks, I'd rather we had taken the direction of some precedent:

My concern with these approaches is that none of them solve the plugin-unhooking-other-plugin problem. Anything that isn't a static identifier is going to fail for that use case. This is a pretty significant use case in PHP-land; I don't see that there are any fundamental differences that would change that in JS-land.

The PHP-land solution is that you always hook something that's globally-exposed, whether that's a function name (always global) or an object method (where you need to expose the object yourself). The problem in JS-land is that functions are not usually globally-exposed, or are completely anonymous, which means we have basically no ID to use. (Anonymous functions for PHP hooks are already discouraged as a best practice for exactly this reason.)

aduth:
>My concern with these approaches is that none of them solve the plugin-unhooking-other-plugin problem.

You'll have to forgive my ignorance, but this is not a pattern I've seen very often. Can you point to a few examples so I have a better idea of their use-case?

I'm not anti-ad-hoc-removal (well, maybe a bit 😄 ), but it's a decision of best compromise: Would I sacrifice the ability to remove other plugin's hooks if it meant I'd only need to provide a hook name and callback to register my own, rather than hook name, vendor name, namespace, callback name, and callback? For me, yes I would.

adamsilverstein:
@pento what would you think about merging as an experimental API, similar to what we did with Heartbeat?

adamsilverstein:
> You'll have to forgive my ignorance, but this is not a pattern I've seen very often. Can you point to a few examples so I have a better idea of their use-case?

An example would be you have installed some plugin that filters post content in some way and you want to use the plugin, but remove its content filtering, perhaps in a certain context only.

pento:
> what would you think about merging as an experimental API, similar to what we did with Heartbeat?

I'm exceedingly skeptical of that option. Heartbeat being merged as an experimental API didn't really achieve any of its goals: it didn't gain much adoption, the adoption it did gain meant we struggled to break backwards compatibility, and it didn't evolve much in subsequent releases.

I think the most successful new API in recent years was the REST API core - it was solid, having been used as the base for the REST API endpoints, there have been lots of examples of plugins building new endpoints based on it, and we haven't had huge issues with needing to break backwards compatibility.

aaronjorbin:
> > My concern with these approaches is that none of them solve the plugin-unhooking-other-plugin problem.

>You'll have to forgive my ignorance, but this is not a pattern I've seen very often. Can you point to a few examples so I have a better idea of their use-case?

A quick grep of the plugin repo shows many examples.  Here is one where a plugin looks to see if woocomerce breadcrumbs are enabled and if so, replaces the action with one of it's own: https://plugins.trac.wordpress.org/browser/catalyst-connect/trunk/inc/frontend.php#L104

Here is an example of modifying the genesis theme: https://plugins.trac.wordpress.org/browser/thememix-pro-genesis/trunk/page-templates/page-login.php#L14

This technique was also talked about as a part of how NPR built the Argo network: https://wordpress.tv/2011/09/25/marc-lavallee-wes-lindamood-plugins-are-blueprints/

aaronjorbin:
It appears the goal is to turn hook names into PHP namespaces.  If so, I think we need about 7 more slashes to really replicate the experience. 

As with all large API changes (Gutenberg, REST-API, etc), this should have a long period of relative stability with solid documentation in order to develop maturity before it is added to core.

adamsilverstein:
> As with all large API changes (Gutenberg, REST-API, etc), this should have a long period of relative stability with solid documentation in order to develop maturity before it is added to core.

Worth noting that the code in this package was originally provided on https://core.trac.wordpress.org/ticket/21170 over 5 years ago. Although we have added some tests and small improvements in the last few months, the core of the library remains unchanged. 

In regards to documentation, here is what we have so far: https://github.com/adamsilverstein/packages/blob/45747e5ad147a6b35f2b7bb508eb139a475e2094/packages/hooks/README.md - this could definitely be improved with additional details and examples.

> This needs more than just time to soak in trunk before 4.9, it's a huge new API.
> Please consider this a blocker: before it's merged, it needs to be integrated into a substantial project to help iron out practical problems that are likely to come up.

Really? This seems pretty solid and well tested to me, and worth having in core as we explore ways to make our JavaScript more extensible, for example the media manager. I am ready to commit this code on https://core.trac.wordpress.org/ticket/21170 - it is mainly 5+ year old code that has had review from project leads and some of our top core JS engineers. This is a solid, reliable hooks library I would use in production today. Sure, we may still find edge cases bugs: let's start using it to find out.

Lets release this as an NPM package, then projects like Gutenberg or WordPress core itself can test it out by importing it from npm.

azaozz:
Argh, the context from https://core.trac.wordpress.org/ticket/21170 is so much missing here :(

Not sure if this is a "big new API", it is a pub/sub that replaces `jQuery( document ).trigger()`. As @adamsilverstein mentions above, it has been waiting in 21170 for over 5 years already. Don't think it can be compared to the REST API both as size and as complexity.

Heartbeat was marked "experimental" not because it needed to "wait" for plugins to start using it. There were two main reasons:
- Some people were concerned that a request every 30 seconds was too much for many hosts to handle. (Btw I'd really want to see what would happen when the REST API gets a decent amount of usage on all installs).
- The browser support for detecting when the user has switched tabs, focused an iframe, or moved away/became inactive was sketchy, so there were several workarounds that worked in different browsers to different degree.

Think it's about time to bite the bullet and finally commit this. The alternative is to continue using `jQuery( document ).trigger()` in core, no better in any way. Also, five years is a very very long time no matter how you look at it :)

adamsilverstein:
wp.hooks has landed in core, let's get this package released to npm as 1.0.0.

https://core.trac.wordpress.org/changeset/41375

adamsilverstein:
Let's get this published on npm as `0.1.0-beta.0`, that way we can try using via import in Gutenberg and WordPress core.

PR #16, opened 2017-07-20
Fix initialization from a fresh repo clone

nylen:
Currently, cloning the repo and attempting to run the tests gives the following error:

```
 FAIL  packages/hooks/src/test/index.test.js
  ● Test suite failed to run

    /home/james/wp/packages/packages/hooks/src/test/index.test.js:4
    import {
    ^^^^^^
    SyntaxError: Unexpected token import
```

This is because `babel-jest` is not installed.  From https://facebook.github.io/jest/docs/getting-started.html#using-babel:

> To use Babel, install the `babel-jest` and `regenerator-runtime` packages [...] Note: Explicitly installing regenerator-runtime is not needed if you use npm 3 or 4 or Yarn

Related to that last point, also adds a `.nvmrc` encouraging use of Node 8 and npm 5, and the `package-lock.json` file from `npm 5` to guarantee the same versions of dependencies for everyone who uses npm 5.

Finally, fixes an issue where `npm run build` would also create a bunch of built `*.test.js` files, which would then be executed on subsequent `npm test` runs.


ntwb:
I've no issues with the changes here though whist testing based on the original report I didn't encounter any of the issues mentioned, here's my terminal output for reference:

```shell
~/dev/tmp 
❯ git clone git@github.com:WordPress/packages.git tmp-packages
Cloning into 'tmp-packages'...
remote: Counting objects: 266, done.
remote: Compressing objects: 100% (37/37), done.
remote: Total 266 (delta 16), reused 33 (delta 13), pack-reused 214
Receiving objects: 100% (266/266), 121.57 KiB | 0 bytes/s, done.
Resolving deltas: 100% (100/100), done.

~/dev/tmp 
❯ cd tmp-packages/

~/dev/tmp/tmp-packages master 
❯ npm i
npm notice created a lockfile as package-lock.json. You should commit this file.
npm WARN tmp-packages No repository field.
npm WARN tmp-packages No license field.

added 573 packages in 9.968s

~/dev/tmp/tmp-packages master* 
❯ npm test

> @ test /Users/netweb/dev/tmp/tmp-packages
> jest

 PASS  packages/url/src/test/index.test.js
  addQueryArgs
    ✓ should append args to an URL without query string (5ms)
    ✓ should append args to an URL with query string (1ms)

Test Suites: 1 passed, 1 total
Tests:       2 passed, 2 total
Snapshots:   0 total
Time:        1.384s
Ran all test suites.

~/dev/tmp/tmp-packages master* 
❯ npm run build

> @ build /Users/netweb/dev/tmp/tmp-packages
> node ./scripts/build.js

>> Building packages 
url
 DONE 
```

ntwb:
I also added the `package-lock.json` file for the `packages/url` component in https://github.com/WordPress/packages/pull/16/commits/24b1dc04cc9b62596ec6abc3aedd255fcb3841ff

.nvmrc
```diff
@@ -0,0 +1 @@
+v8.2.0
```
ntwb: This should just be `8` should it not? That way the latest 8.x.x version is used by way of `nvm use 8`

nylen:
**If we use a specific version of node:** We have to upgrade it manually when we are ready.  This is the approach taken by other projects (for example, WordPress core, which in practice never gets updated, and Calypso, which has pretty frequent Node version updates).

**If we use the latest version:**  Upgrades can break things without it being clear what's going on.  (This will happen anyway on Travis builds.  Maybe it is better to be consistent with what Travis is doing, one way or the other.)

Also, changing `.nvmrc` to `v8` doesn't ensure that everyone will get the latest version - in my testing, a `nvm install v8` is still required to get an update.

I don't have a strong preference here, but wanted to point out the drawbacks of not pinning to a specific version.

---

I'm no longer seeing the `babel-jest` issue (on a fresh repo clone, it is installed correctly).  I've reverted that change in 4864ba5.

`babel-jest` is a [dependency of `jest-runtime`](https://github.com/facebook/jest/blob/v20.0.4/packages/jest-runtime/package.json#L12) which is a [dependency of `jest-cli`](https://github.com/facebook/jest/blob/v20.0.4/packages/jest-cli/package.json#L25) which is a [dependency of `jest`](https://github.com/facebook/jest/blob/v20.0.4/packages/jest/package.json#L7), so it should be installed normally.  I am thinking this was a temporary bug related to `npm` cache or something.  (I only saw this bug on npm 3.x.)

ntwb:
> I only saw this bug on npm 3.x

😂

ntwb:
I dropped the `8` vs `8.2.0` into this weeks Slack chat:
> https://wordpress.slack.com/archives/C5UNMSU4R/p1500990356596185

tl;dr: Everyone loves Node.js 8.x and cannot wait for its LTS release in October

That said, the general consensus is also around using the _latest_ release though concerns around _upgrading code to use the latest release_ was also raised. I don't see that as an issue as we are also targeting the current 6.x LTS for compat with WP Core.

I think we go with `8` in `.nvmrc`, that should use the latest version of Node.js 8.x that a user has installed locally via `nvm install 8`

Note: I'm suggesting `8` here and not `v8`, `8` is a supported alias of nvm for the latest 8.x branch, I'm not sure what results occur using `v8`

aduth:
It's also possible to define "latest LTS" as an option via `lts/*`:

https://github.com/creationix/nvm#long-term-support

Which could still fall victim to breaking changes, but not so much as "latest version".

nylen:
> Note: I'm suggesting `8` here and not `v8`

This works the same.  From `nvm --help`:

> Note: <version> refers to any version-like string nvm understands. This includes:
>  - full or partial version numbers, starting with an optional "v" (0.10, v0.1.2, v1)

It's pretty natural to use the `v` prefix, e.g. `node --version > .nvmrc`, and this also provides an indicator that it's a version number rather than "here's a random file with just an 8 inside of it".

nylen:
Build is failing due to coverage integration.  Fixing at #22 then will update here.

PR #18, opened 2017-07-24
Add test coverage and run tests during Travis build

nylen:
This PR adds code coverage reporting using our existing `jest` test suite and sends the data to [coveralls.io](http://coveralls.io/).

The initial `.travis.yml` file added in #5 contained `script: true` which basically causes Travis to do nothing.  However, now that we have a working test suite, we should actually run it on Travis builds.


nylen:
For some reason the Travis notifications are not showing up here.  However, I'm going to merge this as it will help with development of #13.

PR #21, opened 2017-07-25
URL: Adding a README for the URL Package

youknowriad:
Everything is in the title


packages/url/README.md
```diff
@@ -0,0 +1,20 @@
+# WordPress Url Package
+
+A collection of utilities to manipulate URLs
+
+## Installation
+
+Install the module
+
+```bash
+npm install @wordpress/url --save
```
youknowriad: I guess the name here is not defined yet, is it?

packages/url/README.md
```diff
@@ -0,0 +1,20 @@
+# WordPress Url Package
```
ntwb: Capitalized `URL`

packages/url/README.md
```diff
@@ -0,0 +1,20 @@
...
+npm install @wordpress/url --save
+```
+
+## Usage
+
+```JS
+import { addQueryArgs } from '@wordpress/url';
+
+// Appends arguments to the query string of a given url
+const newUrl = addQueryArgs( 'https://google.com', { q: 'test' } ); // https://www.google.dz/?q=test
```
ntwb: The line comment here should also use `google.com` as per the example code.

youknowriad: 🙃

packages/url/README.md
aduth: Not yet, but this is what it'd be published as.

packages/url/README.md
```diff
@@ -0,0 +1,20 @@
...
+npm install @wordpress/url --save
+```
+
+## Usage
+
+```JS
+import { addQueryArgs } from '@wordpress/url';
+
+// Appends arguments to the query string of a given url
+const newUrl = addQueryArgs( 'https://google.com', { q: 'test' } ); // https://www.google.com/?q=test
```
aduth: No `www` per example

ntwb: Ha, I missed that 😉

PR #22, opened 2017-07-27
Switch code coverage from coveralls to codecov

nylen:
Same as https://github.com/WordPress/gutenberg/pull/2054.


nylen:
Back to the default codecov settings here.  See https://github.com/WordPress/gutenberg/pull/2055#issuecomment-318380807.

PR #33, opened 2017-08-15
Add test for making sure `addQueryArgs` updates existing arguments.

schlessera:
This is to ensure it will not leave duplicate arguments behind.


omarreiss:
Thanks!

PR #34, opened 2017-08-15
tests: Use the generic domain `example.com` in tests

ntwb:


packages/url/src/test/index.test.js
```diff
@@ -5,23 +5,23 @@ import { addQueryArgs } from '../';
 
 describe( 'addQueryArgs', () => {
 	test( 'should append args to an URL without query string', () => {
-		const url = 'https://andalouses.com/beach';
```
youknowriad: 😭 Just wanted to leave some of "my culture" in there (Not the domain, just the "andalouses" name) which is the name of beach not far from my place.

But don't worry I don't care that much

ntwb: Hehehe, I guessed it was something local to you 🏖

ntwb: p.s. andalouses.com should get a SSL certificate from Lets Encrypt 🔑

youknowriad: Actually didn't open the page :)

aduth: A compromise could be a customized domain which is guaranteed not to resolve to a remote resource, like `andalouses.example`.

See: __2. TLDs for Testing, & Documentation Examples__

http://www.ietf.org/rfc/rfc2606.txt

> To safely satisfy these needs, four domain names are reserved as
   listed and described below.
>
>                   .test
>                .example
>                .invalid
>              .localhost
>
>      ".test" is recommended for use in testing of current or new DNS
>      related code.
>
>      ".example" is recommended for use in documentation or as examples.
>
>      ".invalid" is intended for use in online construction of domain
>      names that are sure to be invalid and which it is obvious at a
>      glance are invalid.
>
>      The ".localhost" TLD has traditionally been statically defined in
>      host DNS implementations as having an A record pointing to the
>      loop back IP address and is reserved for such use.  Any other use
>      would conflict with widely deployed code which assumes this use.

Not sure how much we care, but agree it shouldn't be a URL that's not within control of the WordPress Foundation.

youknowriad: Yeah, don't worry too much. Just use any random test domain

ntwb: Done ❤️

PR #37, opened 2017-08-17
Add a webpack based build to build wp.{moduleName} namespaced objects

adamsilverstein:
Ultimately this build process should probably live in core. Adding this here and removing it from the hooks PR to keep it separate.


webpack.config.js
```diff
@@ -0,0 +1,47 @@
+const entryPointNames = [
```
aduth: Seems like this would become a pain to maintain. Can we use [`fs.readDirSync( './packages' )`](https://nodejs.org/api/fs.html#fs_fs_readdirsync_path_options) instead?

webpack.config.js
```diff
@@ -0,0 +1,47 @@
...
+	'url',
+	'a11y',
+	'dom-ready',
+];
+
+const externals = {
```
aduth: How are we expecting this set to be maintained? None of the included packages in this repository use these globals.

webpack.config.js
```diff
@@ -0,0 +1,47 @@
...
+};
+
+entryPointNames.forEach( entryPointName => {
+	externals[ entryPointName ] = {
+		'this': [ 'wp', entryPointName ],
+	};
+} );
+
+const config = entryPointNames.reduce( ( memo, entryPointName ) => {
+		memo[ entryPointName ] = './packages/' + entryPointName + '/src/index.js';
```
aduth: Extra leading whitespace.

webpack.config.js
```diff
@@ -0,0 +1,47 @@
...
+		return memo;
+ }, {} );
```
aduth: Extra leading whitespace.

package.json
```diff
@@ -2,14 +2,17 @@
   "devDependencies": {
     "babel-core": "^6.25.0",
     "babel-plugin-transform-object-rest-spread": "^6.23.0",
+    "babel-plugin-transform-runtime": "^6.23.0",
```
aduth: Is this needed? Where is it used?

webpack.config.js
youknowriad: There's a `getAllPackages` and `getPackageNameFromFile` functions in the `scripts/build.js` file. Could be extracted into a helper file.

youknowriad:
Can I ask why is this included here? Shouldn't this be in Core instead? importing the modules using npm and creating the globals there?

adamsilverstein:
Closing in favor of building these packages on the core side now that we have webpack in core.

PR #39, opened 2017-09-27
Correctly enable dashes in regexes in validateNamespace

adamsilverstein:


youknowriad:
Thanks for this fix, can we add a test covering this bug?

adamsilverstein:
> Thanks for this fix, can we add a test covering this bug?

Yep, will do.

adamsilverstein:
@youknowriad added a test... before the regex change, the test fails.

PR #40, opened 2017-09-27
Add Hooks to objects via composition or as a mixin

adamsilverstein:
This PR is an experiment and seeks to encapsulate logic on a per object basis and change hooks from a global event handler to a mixin added to objects via composition.

Hooks can be added to an object via composition:

`myObject.hooks = new Hooks()`

API functions can then be called on the object: `myObject.hooks.addAction()`,  `myObject.hooks.applyFilters()`,etc.

This removes the requirement for complex namespaces when adding hooks, we only need a functionName to ensure we can always remove a hook. As @westonruter pointed out in [his comment on the trac ticket
](https://core.trac.wordpress.org/ticket/21170#comment:180) this offers several potential advantages. we could add hooks in core at any level we felt appropriate, for example `wp.media.hooks` or - if we decided it was appropriate - right at `wp.hooks`. 

Are we loosing anything other than potentially reduced file sizes by switching this from a global to a per object approach?


packages/hooks/README.md
```diff
@@ -1,34 +1,28 @@
...
-* `doingAction( 'hook' )`
-* `doingFilter( 'hook' )`
-* `didAction( 'hook' )`
-* `didFilter( 'hook' )`
-* `hasAction( 'hook' )`
-* `hasFilter( 'hook' )`
-
+Hooks can be added to an object via composition:
+
+`myObject.hooks = new Hooks()`
```
westonruter: In this case, should the `Hooks` function take an optional `context` arg which is then used when running callbacks? For example:

```js
myObject.hooks = new Hooks( myObject );
```

With the parent object then defined, then it would be available for use as `this` context. If not provided, then `this` would be the instance of `hooks` itself.

westonruter: Also, `Hooks` could be used as a mixin:

```js
Object.assign( myObject, new Hooks() );
```

Usage:

```js
myObject.addFilter( … );
```

adamsilverstein: good suggestion, I'll add and see if I can add tests.

adamsilverstein: Right, I'll document and test this as well

adamsilverstein:
Appreciate any feedback from @aduth, @omarreiss or @youknowriad (or anyone else!) on the approach here.

packages/hooks/src/index.js
```diff
@@ -7,34 +6,30 @@ import createCurrentHook from './createCurrentHook';
...
+		'removeAllFilters': createRemoveHook( HOOKS.filters, true ),
+		'doAction':         createRunHook( HOOKS.actions ),
+		'applyFilters':     createRunHook( HOOKS.filters, true ),
+		'currentAction':    createCurrentHook( HOOKS.actions ),
+		'currentFilter':    createCurrentHook( HOOKS.filters ),
+		'doingAction':      createDoingHook( HOOKS.actions ),
+		'doingFilter':      createDoingHook( HOOKS.filters ),
+		'didAction':        createDidHook( HOOKS.actions ),
+		'didFilter':        createDidHook( HOOKS.filters ),
+		'HOOKS':            HOOKS,
```
youknowriad: minor: we could remove the quotes from the keys here.

packages/hooks/README.md
youknowriad: Should we encourage the use of the context? One of the most controversial JavaScript features. What benefit this would give us?

Also what about `createHooks` instead of `new Hooks()`?

packages/hooks/src/index.js
adamsilverstein: ah, right. will fix.

packages/hooks/README.md
adamsilverstein: The benefit of context I see is that inside your callbacks you could easily refer to the calling object which seems like a common use case, otherwise you would have to bind 'this' to the callback or hoist it to get a handle.

aduth: >otherwise you would have to bind 'this' to the callback or hoist it to get a handle

An arrow function would work as well, which is not outlandish to target in the near future even without a build step (Vue 3+, for example, is [dropping support for non-ES2015-capable browsers](https://hashnode.com/ama/with-vuejs-team-cj7itlrki03ae62wuv2r2005s#cj7k5nd67003k7qwt1147975x)).

packages/hooks/src/index.js
adamsilverstein: addressed in https://github.com/WordPress/packages/pull/40/commits/031fdfddb6484329c6d884edd8440de69a256e58

adamsilverstein:
@youknowriad swtiched to createHooks naming in https://github.com/WordPress/packages/pull/40/commits/b8d5f287b56f8bc056ee76e221611620f2da4da9

packages/hooks/src/index.js
```diff
@@ -7,34 +6,30 @@ import createCurrentHook from './createCurrentHook';
...
-export const currentFilter = createCurrentHook( HOOKS.filters );
-
-// Doing action/filter: true while a hook is being run.
-export const doingAction = createDoingHook( HOOKS.actions );
-export const doingFilter = createDoingHook( HOOKS.filters );
-
-// Did action/filter functions.
-export const didAction = createDidHook( HOOKS.actions );
-export const didFilter = createDidHook( HOOKS.filters );
+export const createHooks = () => {
```
youknowriad: Should this be `export default`?

packages/hooks/README.md
```diff
@@ -1,34 +1,35 @@
...
+* `didAction( 'hookName' )`
+* `didFilter( 'hookName' )`
+* `hasAction( 'hookName' )`
+* `hasFilter( 'hookName' )`
 
-### Background
-See ticket [#21170](http://core.trac.wordpress.org/ticket/21170) for more information.
+Hooks can be added to an object via composition:
+
+`myObject.hooks = new Hooks( myObject )`
```
youknowriad: We should update the docs after the recent changes

packages/hooks/src/index.js
```diff
@@ -7,34 +6,30 @@ import createCurrentHook from './createCurrentHook';
...
+		removeAllFilters: createRemoveHook( HOOKS.filters, true ),
+		doAction:         createRunHook( HOOKS.actions ),
+		applyFilters:     createRunHook( HOOKS.filters, true ),
+		currentAction:    createCurrentHook( HOOKS.actions ),
+		currentFilter:    createCurrentHook( HOOKS.filters ),
+		doingAction:      createDoingHook( HOOKS.actions ),
+		doingFilter:      createDoingHook( HOOKS.filters ),
+		didAction:        createDidHook( HOOKS.actions ),
+		didFilter:        createDidHook( HOOKS.filters ),
+		HOOKS:            HOOKS,
```
aduth: In the context of a singular constant object, the capitalized variable made more sense. Here since it's exposed as a property of the object we're creating, I think we ought to just lower-case it.

adamsilverstein: Good suggestion. 

I was able to remove the HOOKS abstraction entirely in https://github.com/WordPress/packages/pull/40/commits/4cf37499bb4fb8fd6ae03d54a228164d708591da; since we return a hooks object now, we can track filters and actions directly on it instead of adding these as sub-objects to an additional HOOKS object.

packages/hooks/src/index.js
adamsilverstein: updated in https://github.com/WordPress/packages/pull/40/commits/5a6a91887c2a98a0f1fcb9a9ca1477ef0093c0d2

packages/hooks/README.md
adamsilverstein: updated in https://github.com/WordPress/packages/pull/40/commits/f1c4f1fa0d8760e72d93604292e3466b9d2716c5

adamsilverstein:
I'd like to get this merged/deployed so we can get this to npm.   cc: @westonruter @omarreiss @youknowriad @aduth

packages/hooks/src/validateNamespace.js
```diff
@@ -13,13 +13,8 @@ function validateNamespace( namespace ) {
 		return false;
 	}
 
-	if ( ! /^[a-zA-Z][a-zA-Z0-9_.\-/]*$/.test( namespace ) ) {
```
aduth: The documentation for this function still reflects the `vendor/plugin/function` format. It's also a bit of a misnomer to call this a namespace validation, since at this point it's really just checking that the function name takes a particular form (there is no namespace).

PR #42, opened 2017-10-03
a11y: Use screen-reader-text styles for speak container

### LINKED ISSUE #38: a11y: Speak region causes page overflow
The `@wordpress/a11y` module adds an element container to the page which acts as a target for `speak` text. However, its current styling will cause overflow (scroll) on the page when inserted as a sibling of `#wpwrap`.

https://github.com/WordPress/packages/blob/fd7fff1458f740204cc4e4ee2dfdb804a1c8c38b/packages/a11y/src/addContainer.js

__Steps to reproduce:__

1. With the [Gutenberg](https://wordpress.org/plugins/gutenberg/) plugin installed, navigate to Gutenberg > New Post in the admin dashboard
2. Open your developer tools console
3. Remove all nodes except `#wpwrap` and `.a11y-speak-region`
   - `[ ...document.body.querySelectorAll( ':not( #wpwrap ):not( .a11y-speak-region )' ) ].forEach( ( node ) => node.parentNode.removeChild( node ) );`

__Expected:__ No scroll on page
__Actual:__ Scroll on page

We should seek to either:

- Append `.a11y-speak-region` within `#wpwrap`
- Style `.a11y-speak-region` in such a way to not occupy any space

cc @omarreiss

aduth:
Fixes #38 

This pull request seeks to resolve the a11y speak container styling issues noted at #38. Specifically, the updated styles help avoid a scroll overflow issue, doing so by copying the `.screen-reader-text` styles used elsewhere in the WordPress codebase:

https://github.com/WordPress/WordPress/blob/4.8.2/wp-admin/css/common.css#L120-L133

__Open questions:__

After publishing this branch, I noted that the styles have since been updated in the trunk branch to use modern CSS properties. We may consider adopting those styles here (cc @afercia).

https://github.com/WordPress/WordPress/commit/c65fe27ba10d7ecc97d0ef5cebd690878843a558

__Testing instructions:__

The original issue was observed in overflow apparent in Gutenberg prior to the workaround put in place in WordPress/gutenberg@0f77b9cf988a76e449434db357558dab261a4039 . Verify that prior to this commit, with these changes, no overflow exists:

1. After checking out this branch, create an npm symlink for the `a11y` package:
   - `cd packages/a11y && npm ln`
2. In Gutenberg, check out a commit prior to the workaround fix:
   - `git checkout 04727b731a2ec307c9bad1d9d239d3c5e01f89a4`
3. In Gutenberg, establish a link to the local `@wordpress/a11y` branch:
   - `npm ln @wordpress/a11y`
4. In Gutenberg, start the local development build
   - `npm run dev`
5. Navigate to the local WordPress installation running your copy of Gutenberg and verify that there is no scroll overflow on the Gutenberg editor screens

### DIFF
--- a/packages/a11y/src/addContainer.js
+++ b/packages/a11y/src/addContainer.js
@@ -12,8 +12,17 @@ const addContainer = function( ariaLive ) {
 	container.id = 'a11y-speak-' + ariaLive;
 	container.className = 'a11y-speak-region';
 
-	let screenReaderTextStyle = 'clip: rect(1px, 1px, 1px, 1px); position: absolute; height: 1px; width: 1px; overflow: hidden; word-wrap: normal;';
-	container.setAttribute( 'style', screenReaderTextStyle );
+	container.setAttribute( 'style', (
+		'position: absolute;' +
+		'margin: -1px;' +
+		'padding: 0;' +
+		'height: 1px;' +
+		'width: 1px;' +
+		'overflow: hidden;' +
+		'clip: rect( 0 0 0 0 );' +
+		'border: 0;' +
+		'word-wrap: normal !important;'
+	) );
 	container.setAttribute( 'aria-live', ariaLive );
 	container.setAttribute( 'aria-relevant', 'additions text' );
 	container.setAttribute( 'aria-atomic', 'true' );
--- a/packages/a11y/src/test/addContainer.test.js
+++ b/packages/a11y/src/test/addContainer.test.js
@@ -8,7 +8,7 @@ describe( 'addContainer', () => {
 			expect( container ).not.toBe( null );
 			expect( container.className ).toBe( 'a11y-speak-region' );
 			expect( container.id ).toBe( 'a11y-speak-polite' );
-			expect( container.getAttribute( 'style' ) ).toBe( 'clip: rect(1px, 1px, 1px, 1px); position: absolute; height: 1px; width: 1px; overflow: hidden; word-wrap: normal;' );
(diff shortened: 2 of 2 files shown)


afercia:
We should discuss a bit `'clip: rect( 0 0 0 0 );'` this is an old syntax used to support old browsers. The correct syntax now is `rect(1px, 1px, 1px, 1px);`. Also, `clip` is deprecated.

Also, the WordPress `screen-reader-text` CSS class has been updated very recently, see https://core.trac.wordpress.org/changeset/41622

Worth also noting there's a Safari / VoiceOver (pre-existing) bug triggered by the absolute positioning, see https://core.trac.wordpress.org/ticket/42006

aduth:
Thanks for the feedback @afercia . https://github.com/h5bp/html5-boilerplate/issues/1985 is a rollercoaster of an issue 😄  It doesn't appear there's a good solid fix here yet? Not sure how we'd want to move forward, if the changes from https://github.com/WordPress/WordPress/commit/c65fe27ba10d7ecc97d0ef5cebd690878843a558 would be sufficient as an interim improvement.

afercia:
@aduth unfortunately the recent changes don't fix the issue. Better luck if anyone has friends at Apple maybe 🙂 as the wrong order issue is really a Safari / VoiceOver bug.
Yep, seems at boilerplate they've tried a lot of things and then reverted .

omarreiss:
I think this is an improvement in itself. Aligning with recent screen reader text styling update can be done in a separate pull. Merging.

PR #44, opened 2017-11-24
Add option to specify namespace validation regex

tg-ephox:
Allows hook namespace to match Gutenberg block registrations.

See https://github.com/WordPress/gutenberg/pull/3621

### DIFF
--- a/packages/hooks/src/createAddHook.js
+++ b/packages/hooks/src/createAddHook.js
@@ -8,7 +8,7 @@ import validateHookName from './validateHookName.js';
  *
  * @return {Function}       Function that adds a new hook.
  */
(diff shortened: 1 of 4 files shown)


aduth:
I think the `\` will inevitably cause trouble elsewhere as it did in Gutenberg and that the default should be changed to `/`. I don't think we need to rush to add support for specifying a custom validation expression.

adamsilverstein:
@tg-ephox thanks for the PR - The goal of the check is to enforce consistent naming conventions so rather than make this completely flexible, we are going to adjust the namespace requirements to match what is used in gutenberg, and avoid the problematic `\`  in https://github.com/WordPress/packages/pull/47.

PR #45, opened 2017-11-24
Hooks: Update hooks public API to make it possible to apply to `wp.hooks` directly

gziolo:
We discussed in during the last Core JS meeting. See: https://wordpress.slack.com/archives/C5UNMSU4R/p1511276154000233.

The idea here is to have `@wordpress/hooks` === `wp.hooks`, so we could use it directly in Gutenberg. It was raised while working on hooks integration with Gutenberg: https://github.com/WordPress/gutenberg/pull/3493. The existing implementation created an ambiguity where we weren't able to easily expose hooks as `wp.hooks` and import using `@wordpress/hooks` at the same time. This PR opens this possibility by making import statement for `@wordpress/hooks` identical to what is going to be exposed as `wp.hooks`:
```js
wp.hooks = require( '@wordpress/hooks' );
```


gziolo:
> The question becomes: When would a developer consuming `@wordpress/hooks` want to create their own registry of hooks via `createHooks` ? Would it be bad for many independent modules who each depend on `@wordpress/hooks` to share the same registry? If 

I think that the currently proposed implementation in Gutenberg uses both approaches. We use `createHooks` to isolate all filters related to blocks and the global instance of hooks for the editor. I'm still not convinced if that is the best way to move forward, but it aligns with the API we expose :)

`wp.blocks.registerBlock` + `wp.blocks.addFilter` vs `wp.hooks.addFilter`.

PR #47, opened 2017-11-28
Hooks: Allow slashes in the namespace

gziolo:
We would like to use slashes in the hook's namespace as follows:
```js
addFilter( 'registerBlockType', 'core/custom-class-name/attribute', addAttribute );
addFilter( 'BlockEdit', 'core/custom-class-name/inspector-control', addInspectorControl );
addFilter( 'getSaveContent.extraProps', 'core/custom-class-name/save-props', addSaveProps );
```

It is not possible at the moment. This PR changes it and relaxed validation to allow slashes, too.


adamsilverstein:
> Should we disallow backslash \ for the issues it had presented in behaving as part of an escape sequence?

I think backslash is disallowed with the current check.

gziolo:
> I think backslash is disallowed with the current check.

I will double check before merging. From my own experience from the past, it is not allowed, but @aduth found some magical use case where it doesn't complain.

packages/hooks/src/validateNamespace.js
```diff
@@ -13,8 +13,8 @@ function validateNamespace( namespace ) {
 		return false;
 	}
 
-	if ( ! /^[a-zA-Z][a-zA-Z0-9_.\-]*$/.test( namespace ) ) {
-		console.error( 'The namespace can only contain numbers, letters, dashes, periods and underscores.' );
+	if ( ! /^[a-zA-Z][a-zA-Z0-9_.\-/]*$/.test( namespace ) ) {
```
aduth: The forward slash needs to be escaped `\/`.

aduth:
>I think backslash is disallowed with the current check.

Yeah, you're right, I was thrown off by the escaping of dash `-`, but also that we already use a `core\foo` convention in Gutenberg? Why would that have been allowed previously, but not changed here?

adamsilverstein:
@aduth are you sure you are pulling in the latest version in Gutenberg with `@next`? Otherwise you may not be using the latest version from npm. Probably time to publish a release :)

packages/hooks/src/test/index.test.js
```diff
@@ -193,6 +193,13 @@ test( 'cannot add filters with invalid namespaces', () => {
 	);
 } );
 
+test( 'cannot add filters with namespace containing backslash', () => {
+	addFilter( 'hook_name', 'i\n\v\a\l\i\d\a\n\a\m\e', () => null );
```
gziolo: Fun fact is that `invalidn\amae` will pass. I have no idea how to fix that. We can open a follow-up if we are really concerned about it.

PR #52, opened 2017-12-15
Feature/events on add remove

adamsilverstein:
* needs unit tests

### DIFF
--- a/packages/hooks/README.md
+++ b/packages/hooks/README.md
@@ -43,9 +43,9 @@ Hooks can be added to an object via composition:
 
 API functions are then be called: `myObject.hooks.addAction()`.
 
-### Events on action/filter add or remove.
+### Events on action/filter add or remove
 
 Whenever an action or filter is added or removed, a matching `hookAdded` or `hookRemoved` action is triggered.
 
-* `hookAdded` is triggered when `hooks.addFilter` or `hooks.addAction` is called, passing values for `hookName`, `functionName`, `callback` and `priority`.
-* `hookRemoved` is triggered when `hooks.removeFilter` or `hooks.removeAction` is called, passing values for `hookName` and `functionName`.
+* `hookAdded` action is triggered when `addFilter()` or `addAction()` method is called, passing values for `hookName`, `functionName`, `callback` and `priority`.
+* `hookRemoved` action is triggered when `removeFilter()` or `removeAction()` method is called, passing values for `hookName` and `functionName`.
--- a/packages/hooks/README.md
+++ b/packages/hooks/README.md
@@ -22,7 +22,7 @@ A lightweight & efficient filter and action manager.
 * `addAction( 'hookName', 'functionName', callback, priority )`
 * `addFilter( 'hookName', 'functionName', callback, priority )`
 * `removeAction( 'hookName', 'functionName' )`
-* `removeFilter( 'hookName',  'functionName' )`
+* `removeFilter( 'hookName', 'functionName' )`
 * `removeAllActions( 'hookName' )`
 * `removeAllFilters( 'hookName' )`
 * `doAction( 'hookName', arg1, arg2, moreArgs, finalArg )`
@@ -41,4 +41,11 @@ Hooks can be added to an object via composition:
 
 `myObject.hooks = createHooks();`
 
-API functions are then be called: `myObject.hooks.addAction()`...
+API functions are then be called: `myObject.hooks.addAction()`.
+
+### Events on action/filter add or remove.
+
+Whenever an action or filter is added or removed, a matching `hookAdded` or `hookRemoved` action is triggered.
+
+* `hookAdded` is triggered when `hooks.addFilter` or `hooks.addAction` is called, passing values for `hookName`, `functionName`, `callback` and `priority`.
+* `hookRemoved` is triggered when `hooks.removeFilter` or `hooks.removeAction` is called, passing values for `hookName` and `functionName`.
--- a/packages/hooks/src/createAddHook.js
+++ b/packages/hooks/src/createAddHook.js
@@ -1,5 +1,6 @@
 import validateNamespace from './validateNamespace.js';
 import validateHookName from './validateHookName.js';
+import { doAction } from './';
 
 /**
  * Returns a function which, when invoked, will add a hook.
@@ -68,6 +69,8 @@ function createAddHook( hooks ) {
 				runs: 0,
 			};
 		}
+
+		doAction( 'hookAdded', hookName, namespace, callback, priority );
 	};
 }
 
--- a/packages/hooks/src/test/index.test.js
+++ b/packages/hooks/src/test/index.test.js
@@ -630,7 +630,7 @@ test( 'Test `this` context via composition', () => {
 } );
 
 // Test adding a hook triggers a hookAdded action.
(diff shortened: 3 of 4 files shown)


adamsilverstein:
Added some tests for these new events.  My only concern here is the potential for recursion (if you add or remove a hook in the add/remove hook handler).

gziolo:
The same issue applies when someone triggers click event inside onClick handler 😃 I will think about it a bit more. We need to document this hookAdded behavior.

adamsilverstein:
Yea, We may not need to account for recursion, maybe warn about it? 👍 to documenting this.

packages/hooks/src/createAddHook.js
```diff
@@ -68,6 +69,8 @@ function createAddHook( hooks ) {
 				runs: 0,
 			};
 		}
+
+		doAction( 'hookAdded', hookName, namespace, callback, priority );
```
gziolo: Should we guard against `'hookName' === 'hookAdded'`? :)

packages/hooks/src/test/index.test.js
```diff
@@ -64,6 +64,14 @@ function action_c() {
 	window.actionValue += 'c';
 }
 
+function handle_hook_added ( ...args ) {
+	window.hookAdded = { ...args };
```
gziolo: I'm sure it could be tested using local scope with `let hookAdded, hookRemoved` defined outside of the functions that use them.

packages/hooks/src/test/index.test.js
```diff
@@ -620,3 +628,30 @@ test( 'Test `this` context via composition', () => {
 	const testObject2 = {};
 	Object.assign( testObject2, createHooks() );
 } );
+
+// Test adding a hook triggers a hookAdded action.
```
gziolo: Those comments for tests are obsolete as they repeat the test description. Here and in the line 646.

packages/hooks/src/test/index.test.js
```diff
@@ -620,3 +628,30 @@ test( 'Test `this` context via composition', () => {
...
+test( 'adding a hook triggers a hookAdded action passing all callback details', () => {
+	const testObject = {};
```
gziolo: `testObject` is never used in this test. The same applies in the line 648.

packages/hooks/src/test/index.test.js
```diff
@@ -620,3 +628,30 @@ test( 'Test `this` context via composition', () => {
...
+
+	addAction( 'hookAdded', 'my_callback', handle_hook_added );
```
gziolo: Another way of testing it is to pass Jest spy instead of `handle_hook_added` callback and verify if spy was called with the same set of data:

```js
const hookAddedSpy = jest.fn();

...

expect( hookAddedSpy ).toHaveBeenCalledWith( 'testHook', 'my_callback2', action_a, 9 );
```

packages/hooks/src/test/index.test.js
gziolo: Another option is described below: https://github.com/WordPress/packages/pull/52/files#r157469346.

gziolo:
We can iterate on tests a bit. We should at least remove unused code and obsolete comments. Let me know what do you think about other test related comments. I can send commit with a proposed refactor. That leaves as with documentation and we are good to go :)

packages/hooks/src/createAddHook.js
adamsilverstein: that is what i was wondering, but i think fine to keep it simple for now

adamsilverstein:
@gziolo thanks for the review, i think your comments all make sense, leaving some comments inline. I'll work on adding documentation.

packages/hooks/src/test/index.test.js
adamsilverstein: 👍 sounds good if you want to change that

packages/hooks/src/test/index.test.js
adamsilverstein: ok, fine to remove.

packages/hooks/src/test/index.test.js
adamsilverstein: right, fine to remove

packages/hooks/src/test/index.test.js
adamsilverstein: nice, that is much cleaner. I haven't don't much test writing in Jest, thanks for pointing that out.

gziolo:
Awesome, I will address my own comments and hopefully merge tomorrow together with your docs changes. Thank you for a great team-work  👍

packages/hooks/src/test/index.test.js
gziolo: No worries. I'm still learning Jest, it has really well-though API.

packages/hooks/src/createAddHook.js
gziolo: Once I added `jest.fn` to spy on it, it revealed that calling:
```js
addAction( 'hookAdded', 'my_callback', hook_added_spy );
```
triggers `hookAdded` action. I'll add condition to prevent that.

adamsilverstein:
I added some additional tests for addAction/removeAction and took a pass at documenting the basics in the Readme. Once travis completes I think this is good to merge.

packages/hooks/README.md
```diff
@@ -41,4 +41,11 @@ Hooks can be added to an object via composition:
 
 `myObject.hooks = createHooks();`
 
-API functions are then be called: `myObject.hooks.addAction()`...
+API functions are then be called: `myObject.hooks.addAction()`.
+
+### Events on action/filter add or remove.
```
gziolo: We don't put `.` at the end of section title in other places.

packages/hooks/README.md
```diff
@@ -41,4 +41,11 @@ Hooks can be added to an object via composition:
...
+
+Whenever an action or filter is added or removed, a matching `hookAdded` or `hookRemoved` action is triggered.
+
+* `hookAdded` is triggered when `hooks.addFilter` or `hooks.addAction` is called, passing values for `hookName`, `functionName`, `callback` and `priority`.
```
gziolo: What about rephrasing to:

> `hookAdded` action is triggered when `addFilter()` or `addAction()` method is called, passing values for `hookName`, `functionName`, `callback` and `priority`.

A similar change could be applied to the 2nd line.

packages/hooks/src/test/index.test.js
```diff
@@ -620,3 +620,58 @@ test( 'Test `this` context via composition', () => {
...
+	// Test addAction.
+	addAction( 'testAction', 'my_callback2', action_a, 9 );
+	expect( hook_added_spy ).toHaveBeenCalledTimes( 1 );
+	expect( hook_added_spy ).toHaveBeenCalledWith(
+		'testAction',
+		'my_callback2',
+		action_a,
+		9
+	);
+
```
gziolo: Personally, I would split this and the following test into 2 parts to make them more focused. However, it's not mandatory.

gziolo:
I left some minor comments, which aren't blockers. Feel free to merge this one whenever you have time 👍

packages/hooks/src/test/index.test.js
adamsilverstein: Yea, making the tests more granular like this is nicer - thanks!

PR #53, opened 2017-12-15
Add package: autop

aduth:
Related: https://github.com/WordPress/gutenberg/pull/4005

This pull request seeks to introduce a new `autop` package with two named exports: `autop` and `removep`. Originally these were extracted verbatim from [core's `editor.js`](https://github.com/WordPress/wordpress-develop/blob/53f1c74e1c55910f9b7e4aebc8033571d3e52df8/src/wp-admin/js/editor.js#L890-L1156). However, it was found that the JavaScript implementation of `autop` was lacking in many ways when I sought to port the equivalent PHP tests for `wpautop`. You can see the failing skipped tests in f31289b.

Instead, I _re-_ ported the function directly from the [PHP implementation](https://github.com/WordPress/wordpress-develop/blob/53f1c74e1c55910f9b7e4aebc8033571d3e52df8/src/wp-includes/formatting.php#L438-L608) as faithfully as possible. With the reimplementation in 4846972, all test cases ported from the PHPUnit tests pass.

__Testing instructions:__

Ensure unit tests pass:

```
npm test
```

### DIFF
--- a/packages/autop/README.md
+++ b/packages/autop/README.md
@@ -1,0 +1,30 @@
+# @wordpress/autop
+
+JavaScript port of WordPress's automatic paragraph function `autop` and the `removep` reverse behavior.
+
+## Installation
+
+Install the module
+
+```bash
+npm install @wordpress/autop --save
+```
+
+### Usage
+
+Import the desired function(s) from `@wordpress/autop`:
+
+```js
+import { autop, removep } from '@wordpress/autop';
+
+autop( 'my text' );
+// "<p>my text</p>"
+
+removep( '<p>my text</p>' );
+// "my text"
+```
+
+### API Usage
+
+* `autop( text: string ): string`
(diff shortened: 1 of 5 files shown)


azaozz:
> However, it was found that the JavaScript implementation of autop was lacking in many ways

Not exactly :)
The JS autop is meant to work together with TinyMCE and the browser auto-correct: white space doesn't matter, closing `</p>` tags are optional, TinyMCE converts text that is directly in the body to paragraphs, etc. The three combined have generally less edge case errors than the PHP autop. Some PHP tests were probably failing as they look also at white space between tags.

Nevertheless this version will also work properly, a bit slower perhaps. Did some "manual" testing in the current editor and didn't see regressions except one case where extra empty paragraphs were added (but couldn't repeat/confirm).

PR #55, opened 2017-12-19
Scripts: New package with scripts for plugins

gziolo:
Still work in progress ...

This implementation is based on the [react-scripts](https://github.com/facebookincubator/create-react-app/tree/master/packages/react-scripts). Grunt tasks are copied directly from `wp scaffold plugin` templates. See [gruntfile](https://github.com/gziolo/scaffold-command/blob/master/templates/plugin-gruntfile.mustache) for reference.

The end goal would be to run it as:
- `wp-scripts i18n --pluginSlug=abc --textDomain=xyz`
- `wp-scritps readme`

To include you would have to include in `devDependencies` only this:
```json
"@wordpress/scripts": "1.0.0"
```

### Disclaimer

I'm not promoting `Grunt` by any means. It was easy to copy what we already have in `wp-cli` to show how it could benefit plugin developers. Script names could be prefixed with `plugin-` in this case to make it more clear where it is applicable. This could also live in `plugin-scripts` package to avoid confusion.


gziolo:
@danielbachhuber, this is the proof of concept I wanted to share with you :)

ntwb:
FYI: WordPress' official build tool is Grunt, and we've recently added Webpack. Official WordPress repos should use Grunt over Gulp, the reason for this is primarily to reduce the required volume of learning for contributors to WordPress projects.

packages/a11y/package.json
```diff
@@ -18,7 +18,7 @@
     "url": "https://github.com/WordPress/packages/issues"
   },
   "dependencies": {
-    "@wordpress/dom-ready": "^0.1.0-beta.6"
+    "@wordpress/dom-ready": "0.1.0-beta.6"
```
ntwb: This change shouldn't be in this PR should it?

ntwb: Likewise the `dom-ready/package-lock.json` and `a11y/package-lock.json` changes shouldn't be in this PR either?

gziolo: Yes, I will remove those changes.

gziolo:
> FYI: WordPress' official build tool is Grunt, and we've recently added Webpack. Official WordPress repos should use Grunt over Gulp, the reason for this is primarily to reduce the required volume of learning for contributors to WordPress projects.

The main goal of this package is to make it an implementation detail if WordPress is using Grunt, Gulp, Webpack or whatever is necessary to make things done. I can envision using all of those tools in this package, as well as having them replaced in the far future with something that is yet to be created. This approach promotes usage of [npm run-script](https://docs.npmjs.com/cli/run-script) which is a well-established pattern in the JS community.

danielbachhuber:
Can I use this _and_ a Gruntfile in my own project?

e.g. this project would expose some `wp-scripts readme` I can run, but then I can continue to use my `grunt styles` for compiling SASS.

package.json
```diff
@@ -25,7 +25,7 @@
   "scripts": {
     "build-clean": "rimraf ./packages/*/build ./packages/*/build-module",
     "build": "node ./scripts/build.js",
-    "postinstall": "lerna bootstrap --hoist && npm run build",
```
aaronjorbin: Is removing `--hoist` intentional?

packages/scripts/Gruntfile.js
```diff
@@ -0,0 +1,62 @@
+module.exports = function( grunt ) {
+
+	'use strict';
+
+	const pkg = grunt.file.readJSON( 'package.json' );
+	const pluginSlug = grunt.option( 'pluginSlug' ) || pkg.name;
+	const textDomain = grunt.option( 'textDomain' ) || '';
```
aaronjorbin: Should we start a standard of allowing package.json to contain a WordPress object that initially has `textDomain` and `pluginSlug` properties in it?   `pluginSlug` could fallback to `pkg.name` if `pkg.wordpress.pluginslug` isn't defined.

packages/scripts/Gruntfile.js
```diff
@@ -0,0 +1,62 @@
...
+
+	console.log( pluginSlug, textDomain );
```
aaronjorbin: Leftover Dubugging code?

packages/scripts/Gruntfile.js
```diff
@@ -0,0 +1,62 @@
...
+			},
+			target: {
+				files: {
+					src: [ '*.php', '**/*.php', '!\.git/**/*', '!bin/**/*', '!node_modules/**/*', '!tests/**/*' ]
+				}
+			}
+		},
+
+		wp_readme_to_markdown: {
+			your_target: {
```
aaronjorbin: This is another area where `pkg.wordpress` could come in handy, since it would allow screenshots to be passed in.

packages/scripts/README.md
```diff
@@ -0,0 +1,15 @@
+# @wordpress/scripts
+
+A collection of JS scripts for WordPress development.
+
+## Installation
+
+Install the module
+
+```bash
+npm install @wordpress/scripts --save
```
aaronjorbin: The PR summary contains the line: "To include you would have to include in devDependencies only this:"

 Would `--saveDev` make more sense?

aaronjorbin:
When this goes is available, let's make sure to announce it on https://make.wordpress.org/core and cross-post it on https://make.wordpress.org/plugins/

package.json
gziolo: Yes, I opened an issue to address it separately #56. I will rebase once it gets merged.

packages/scripts/Gruntfile.js
gziolo: I'm more than happy to do it. I thought it might be too early, but if you raised that now let's go for it 💯

packages/scripts/Gruntfile.js
gziolo: Yes, sir! Removing it :)

gziolo:
@adamsilverstein can we cover it with tests in some simple way?

danielbachhuber:
Probably worth mentioning https://github.com/wp-cli/ideas/issues/68

Sorry I didn't connect the dots before. Given `wp makepot` will soon exist, would it make sense to create a README command too and keep this entirely to PHP?

gziolo:
> Can I use this and a Gruntfile in my own project?
> 
> e.g. this project would expose some wp-scripts readme I can run, but then I can continue to use my grunt styles for compiling SASS.

Yes, it doesn't stop you from having your own Grunt setup in your project. The nice thing about `npm` is that you can use several versions of the same dependency. I guess you would have to set `grunt` as a local dependency and all Grunt plugins to make it work.

packages/scripts/package.json
```diff
@@ -0,0 +1,37 @@
...
+    "bin"
+  ],
+  "bin": {
+    "wp-scripts": "./bin/wp-scripts.js"
+  },
+  "dependencies": {
+    "cross-spawn": "5.1.0",
+    "grunt": "1.0.1",
+    "grunt-wp-i18n": "1.0.1",
+    "grunt-wp-readme-to-markdown": "2.0.1"
```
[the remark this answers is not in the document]
aaronjorbin: @stephenharris is the maintainer. 

In general, I agree that lessening the dependency on Grunt makes sense, but at the same time it's not exposed to the plugin developer, so it is just an implementation detail rather than an api detail.

aaronjorbin:
> Sorry I didn't connect the dots before. Given `wp makepot` will soon exist, would it make sense to create a README command too and keep this entirely to PHP?

I would support this.  If wp-cli can be the tool of truth, I think it makes sense.

gziolo:
Closing in favor of #62.

PR #57, opened 2017-12-27
Element Package: Add the Gutenberg Element Module as a reusable package

youknowriad:
Adds the `element` module extracted as is from "Gutenberg".

If we want to start adding reusable UI components to the packages repository, we need to start by adding the `element` package unless we start using React directly without the abstraction layer.

### DIFF
--- a/packages/element/src/index.js
+++ b/packages/element/src/index.js
@@ -1,0 +1,139 @@
+/**
+ * External dependencies
+ */
+import { createElement, Component, cloneElement, Children } from 'react';
+import { render, findDOMNode, createPortal, unmountComponentAtNode } from 'react-dom';
+import { renderToStaticMarkup } from 'react-dom/server';
+import { camelCase, flowRight, isString, upperFirst } from 'lodash';
+
+/**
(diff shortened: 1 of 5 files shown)


packages/element/src/index.js
```diff
@@ -0,0 +1,139 @@
...
+}
+
+/**
+ * Switches the nodeName of all the elements in the children object
+ *
+ * @param  {?Object} children  Children object
+ * @param  {String}  nodeName  Node name
+ * @return {?Object}           The updated children object
+ */
+export function switchChildrenNodeName( children, nodeName ) {
```
aduth: This doesn't appear to have been used in Gutenberg presently? This seems like the sort of thing that might be the responsibility of the module to implement.

aduth:
Noting that it's proposed to expose `React.Fragment` in https://github.com/WordPress/gutenberg/pull/4139

youknowriad:
As discussed in #core-js meetings, I'm closing this PR. We'll have to revisit at some point when merging all modules.

PR #60, opened 2018-01-08
Jest-console: Add new package with console object matchers for Jest

gziolo:
Testing Matchers for the [Console](https://developer.mozilla.org/en-US/docs/Web/API/Console) object for JavaScript in WordPress.

Moved from Gutenberg project and improved a bit:
- related PR: https://github.com/WordPress/gutenberg/pull/4285
- related issue: https://github.com/WordPress/gutenberg/issues/4251

I integrated it also with the packages repository. It helped me to discover two subtle bugs in the tests that cover `hooks` package.

PR #61, opened 2018-01-09
Enforce using the latest LTS version (8.x) of Node and up

gziolo:
When working on #60 I had some random issues with Travis checks on Node 6. This PR removes support for Node 6 and promotes usage of Node 8.x which is the current active LTS version. I added also an additional check which is executed after `npm install` which ensures that proper versions of Node and npm are installed. This setup matches now what we have in Gutenberg. [Related PR](https://github.com/WordPress/gutenberg/pull/3905).

### Testing

Make sure `npm install` doesn't throw any errors when running Node 8 and npm 5.

Try the same with lower version of Node or npm. You can also temporarily bump versions specified in the `engines` section in `package.json. Make sure error is thrown similar to the following:

```bash
Error: Wanted node version >=9.0.0 (>=9.0.0)
To install node, run `nvm install >=9.0.0` or see https://nodejs.org/
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,11 +1,16 @@
 {
   "license": "GPL-2.0-or-later",
+  "engines": {
+    "node": ">=8.0.0",
+    "npm": ">=5.0.0"
+  },
   "devDependencies": {
     "babel-core": "^6.25.0",
     "babel-loader": "^7.1.1",
     "babel-plugin-transform-object-rest-spread": "^6.23.0",
     "babel-preset-env": "^1.5.2",
     "chalk": "^2.0.1",
+    "check-node-version": "^3.1.1",
     "codecov": "^2.2.0",
     "glob": "^7.1.2",
     "jest": "^22.0.3",
@@ -25,7 +30,7 @@
   "scripts": {
     "build-clean": "rimraf ./packages/*/build ./packages/*/build-module",
     "build": "node ./scripts/build.js",
-    "postinstall": "lerna bootstrap && npm run build",
+    "postinstall": "lerna bootstrap && check-node-version --package && npm run build",
     "test": "jest",
     "test:coverage": "jest --coverage",
     "test:coverage-ci": "jest --coverage && codecov",
--- a/package.json
+++ b/package.json
@@ -29,8 +29,9 @@
   },
   "scripts": {
     "build-clean": "rimraf ./packages/*/build ./packages/*/build-module",
+    "prebuild": "check-node-version --package",
     "build": "node ./scripts/build.js",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -25,7 +30,7 @@
   "scripts": {
     "build-clean": "rimraf ./packages/*/build ./packages/*/build-module",
     "build": "node ./scripts/build.js",
-    "postinstall": "lerna bootstrap && npm run build",
+    "postinstall": "lerna bootstrap && check-node-version --package && npm run build",
```
ntwb: Rather than include `check-node-version --package` in `postinstall` can you add it as `prebuild`, that will ensure it is run prior to `build`, `publish:dev`, and `publish:prod` scripts.

(For those who don't use nvm or for myself where nvm here on my system can be troublesome at times (I use Fish shell rather than Bash) and it doesn't always switch Node.js version automattically)

gziolo: Works for me :)

gziolo: Actually, I will add additional checks to be on the safe side.

gziolo: Addressed with https://github.com/WordPress/packages/pull/61/commits/daf94907760563ab32fa6b01b686e47f928a0ba6.

Adding it as `prebuild` works for all cases, as all of them execute `npm run build`. Great idea 🙇

ntwb:
I fixed a quick typo in https://github.com/WordPress/packages/pull/61/commits/75251a4a5c91edcfbd232dc9166388e4d4ebfa58 and all looks good to me ✅

gziolo:
@ntwb, thanks for review and fixing typo 👋

gziolo:
I don't think that adding `engine` blocks users from consumption, but I might be wrong. According to docs:

> this field is advisory only and will only produce warnings when your package is installed as a dependency.

We can inline those values when calling the command as an alternative approach and skip `engines` field in the `package.json`.

aduth:
No, I wouldn't expect it to block, but sets the expectation that it may not work correctly in older versions; which, do we want to set this expectation?

gziolo:
By the way, the `package.json` from the root folder is never exposed. I assume we are safe. I don't think we put the `engines` config in other `package.json` files :)

ntwb:
I've no concerns is setting that expectation that this repo is Node.js > 8.x LTS. 

----

Looking toward the Gutenberg merge and WordPress 5.0 the build server will be using Node.js 8.9.3
> https://make.wordpress.org/systems/2017/12/11/node-js-8-x-on-build-server/

So, whatever that merge looks like we'll be pretty much exclusively running on Node.js 8.x (nvm on the build server is not an option at this stage on the current infrastructure I'm informed)

PR #65, opened 2018-01-12
chore: Use tabs for indentation in `package.json`, `package-lock.json`, and `lerna.json` files

ntwb:
- `.editorconfig` updated per latest WordPress `.editorconfig` in [#WP43075](https://core.trac.wordpress.org/ticket/43075)

  -  npm 5.x now supports tabbed indentation per http://blog.npmjs.org/post/161081169345/v500 - npm v5.0.0
  -  > _"Indentation is now detected and preserved for `package.json`, `package-lock.json`, and `npm-shrinkwrap.json`. If the package lock is missing, it will default to `package.json`’s current indentation."_

  - `lerna.json` has also been switched to using tabs, this shouldn't be an issue as far as I can see

  - I've also updated autop's `package.json` to match the other packages `package.json` files field structure

  - A new `package-lock.json` file is included for the _hooks_ package as previously it was without a lock file

  - The _url_, _dom-ready_, and _a11y_ packages `package-lock.json` files had out of sync versions compared to their relative `package.json` files, these have also been updated in this PR.


gziolo:
You can still bring changes to autop's `package.json` file :)

ntwb:
This is right for review now 😄 

Checking via https://github.com/WordPress/packages/pull/65/files?w=1 and I noticed that there are also a couple of patch/minor dependency bumps in the `package-lock.json` files that had to regenerated.

ntwb:
> You can still bring changes to autop's `package.json` file from the previous PR :)

Done in https://github.com/WordPress/packages/pull/65/commits/ebd61b633c6835f11058ca4f7ef7aa2ae1450a3c

PR #66, opened 2018-01-21
Babel-preset-default: Add new package containing the default Babel configuration

gziolo:
Follow-up for #62.

This PR extract Babel configuration which is going to be shared between `packages` and `Gutenberg` into its own package.

It should be tested similarly to #62. We need to make sure that all existing commands work as before:
* `npm test`
* `npm run build`

Again, I don't think we need unit tests for this new package as it tests itself because this preset is used by the aforementioned commands.


packages/scripts/package.json
```diff
@@ -30,8 +30,7 @@
 	"devDependencies": {
 		"babel-jest": "^22.0.4",
 		"cross-spawn": "5.1.0",
-		"jest": "^22.0.4",
-		"regenerator-runtime": "^0.11.1"
```
ntwb: Unrelated change, this shouldn't be in this PR, it should be updated by removing it in #62

packages/scripts/scripts/test-script.js
```diff
@@ -1,7 +1,6 @@
 // Do this as the first thing so that any code reading it knows the right env.
 process.env.BABEL_ENV = 'test';
 process.env.NODE_ENV = 'test';
-process.env.PUBLIC_URL = '';
```
ntwb: Unrelated change, this shouldn't be in this PR, it should be updated by removing it in #62

gziolo:
@ntwb good catch, once I move them, we could even rebase it against the `master` and merge before #62 👍

PR #67, opened 2018-01-23
Publish with `next`tag

gziolo:
Executed `npm run publish:dev` to publish new packages with `next` tag:

- @wordpress/a11y@1.0.4
 - @wordpress/autop@1.0.2
 - @wordpress/babel-preset-default@1.0.1-0
 - @wordpress/dom-ready@1.0.1
 - @wordpress/hooks@1.1.2
 - @wordpress/jest-console@1.0.1
 - @wordpress/url@1.0.1


packages/babel-preset-default/package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "@wordpress/babel-preset-default",
-	"version": "1.0.0",
+	"version": "1.0.1-0",
```
ntwb: @gziolo `1.0.1-0`? Was this meant to be `1.0.1`?

gziolo: I wanted to check what happens when i reference this preset from the `package.json` in the root folder. Unfortunately when I do it, it doesn't use the latest code, rather published version from the npm. I will continue using relative path to `./packages/babel-preset-default`. I will publish stable version as `1.0.0`. This was pre-release tag.

ntwb: Right 👍, `.babelrc` isn't taking precedence over `package.json` is it?

gziolo: I don’t remember, they explain it in their docs 😃

PR #68, opened 2018-01-23
build: use `browserslist-config-wordpress` Browserslist shared config

ntwb:
We can use https://www.npmjs.com/package/browserslist-config-wordpress

Repo: https://github.com/ntwb/browserslist-config-wordpress


ntwb:
As a follow-up, we can switch to `@wordpress/browserslist-config` once #69 is merged.

gziolo:
Nice, I wasn't aware it is possible to integrate it this way. We should definitely do that once #69 is moved to WordPress org 👍

ntwb:
Now that #69 has been merged, I've updated the config in https://github.com/WordPress/packages/pull/68/commits/7b5c45074d19a97a54722ef4156e1a538c32e849 to use `@wordpress/browserslist-config`, now to test and see if it all works together 🎉

gziolo:
```Successfully published:
 - @wordpress/babel-preset-default@1.0.1
 - @wordpress/browserslist-config@2.0.0
 - @wordpress/jest-console@1.0.3
 - @wordpress/scripts@1.0.1```

You should be able to reference `@wordpress/browserslist-config` now. I published it :)

**Comment by gziolo** [comment]
@ntwb, can we update and merge this one before we proceed with changes for Gutenberg? I would love to have this one and `@wordpress/jest-preset-default` published so we could test all our recent work in one go.

**Comment by ntwb** [comment]
@gziolo I've resolved the merge conflicts, I'm not sure why Travis CI throws this error:

https://travis-ci.org/WordPress/packages/builds/335101159#L482 
```
Error: Cannot find module '@wordpress/browserslist-config' (While processing preset: "/home/travis/build/WordPress/packages/node_modules/babel-preset-env/lib/index.js")
```

As `@wordpress/browserslist-config` is installed in `dependencies` in `packages/babel-preset-default/package.json` as seen here in this PR https://github.com/WordPress/packages/pull/68/files#diff-eba075d4d802ff8e112e9d92c3036cd6R25

Does a Lerna command need to be run here to link this all together?

File: packages/babel-preset-default/index.js [hunk 1]
```diff
@@ -6,18 +6,7 @@ module.exports = {
...
-					'last 2 Safari versions',
-					'last 2 Edge versions',
-					'last 2 Opera versions',
-					'last 2 iOS versions',
-					'last 1 Android version',
-					'last 1 ChromeAndroid version',
-					'ie 11',
-					'> 1%',
-				],
+				browsers: ['extends @wordpress/browserslist-config'],
```
aduth: Standards: Needs space within the square brackets:

https://make.wordpress.org/core/handbook/best-practices/coding-standards/javascript/#arrays-and-function-calls

Noting that this could serve as interesting evidence against the whitespace exceptions for array and object keys, since it's hard to distinguish between them and arrays proper.

gziolo: We don't have Eslint configured for this repository :(

package.json
```diff
@@ -5,6 +5,7 @@
 		"npm": ">=5.0.0"
 	},
 	"devDependencies": {
+		"@wordpress/browserslist-config": "^2.0.0",
```
gziolo: I'm not happy about that, but this seems to be an issue with the `babel-core` which doesn't follow the following:

> Unfortunately, some tooling does not follow the module resolution spec closely, and instead assumes or requires that dependencies are present specificially in the local node_modules directory. To work around this, it is possible to symlink packages from their hoisted top-level location, to individual package node_modules directory. Lerna does not yet do this automatically, and it is recommended instead to work with tool maintainers to migrate to more compatible patterns.

See more: https://github.com/lerna/lerna/blob/master/doc/hoist.md

This is what I see without this dependency:

> >> Building packages 
a11y
/Users/gziolo/PhpstormProjects/packages/node_modules/babel-core/lib/transformation/file/options/option-manager.js:328
        throw e;
        ^

Error: Cannot find module '@wordpress/browserslist-config' (While processing preset: "/Users/gziolo/PhpstormProjects/packages/node_modules/babel-preset-env/lib/index.js")
    at Function.Module._resolveFilename (module.js:536:15)
    at Function.Module._load (module.js:466:25)
    at Module.require (module.js:579:17)
    at require (internal/module.js:11:18)
    at Object.loadQueries (/Users/gziolo/PhpstormProjects/packages/node_modules/browserslist/node.js:83:19)
    at Function.select (/Users/gziolo/PhpstormProjects/packages/node_modules/browserslist/index.js:632:26)
    at /Users/gziolo/PhpstormProjects/packages/node_modules/browserslist/index.js:151:33
    at Array.reduce (<anonymous>)
    at resolve (/Users/gziolo/PhpstormProjects/packages/node_modules/browserslist/index.js:132:18)
    at browserslist (/Users/gziolo/PhpstormProjects/packages/node_modules/browserslist/index.js:224:16)

gziolo: Hmm, or this is Lerna which doesn't hoist its internal packages? No idea really. @evocateur any ideas why we need to put this dependency also in here?

ntwb: Agreed, less than ideal, not insane, but yeah, having Lerna link these would be 👌

evocateur: Anywhere you need to call a cli in devDependencies, it needs to be specified in the leaf node so lerna knows what to symlink. Hoisting makes it quicker, installing once in the root.

gziolo: Yes, but this one (@wordpress/browserslist-config) is pulled from npm although it is located also inside the packages folder. I hoped to have also the same behavior as in the leaf node for dependencies listed inside the root folder’s package.json. In this case I need to publish package to use the lastest version of it. It makes dogfooding very inconvenient.

PR #69, opened 2018-01-23
feat: Add `@wordpress/browserslist-config` package

ntwb:
This pull request migrates the existing package from:
• npm module: https://www.npmjs.com/package/browserslist-config-wordpress
• GitHub repo: https://github.com/ntwb/browserslist-config-wordpress

to:
• npm module: https://www.npmjs.com/package/@wordpress/browserslist-config
• GitHub repo: https://github.com/WordPress/packages/tree/master/packages/browserslist-config

I'd originally planned on having npm migrate the package for my from my repo/account to @wordpress', though as no other modules are currently consuming this package it was simpler to update and move the code here and have it published as a new package.


gziolo:
Awesome, I’ll take a closer look when I have time 👍

packages/browserslist-config/CHANGELOG.md
```diff
@@ -0,0 +1,17 @@
+# 2.0.0
+
+-   Transfer to @wordpress npm orginization
```
atimmer: `orginization` -> `organization`.

packages/browserslist-config/package.json
```diff
@@ -0,0 +1,33 @@
...
+		"url": "https://github.com/WordPress/packages/issues"
+	},
+	"main": "index.js",
+	"engines": {
+		"node": ">=8"
+	},
+	"devDependencies": {
+		"browserslist": "^2.11.3"
+	},
+	"peerDependencies": {
```
gziolo: Is it necessary since it is already listed as `devDependencies`? Or `devDependencies` should be `dependencies`? I never know :)

packages/browserslist-config/test/index.test.js
```diff
@@ -0,0 +1,17 @@
...
+});
+
+it( 'should export an array', () => {
+	expect( Array.isArray( config ) ).toBe( true );
+});
+
+it( 'should not contain invalid queries', () => {
+	jest.doMock( '@wordpress/browserslist-config', () => require( '../index' ), { virtual: true });
+
+	const result = browserslist(['extends @wordpress/browserslist-config']);
```
gziolo: Spacing should be fixed to match coding styles.

packages/browserslist-config/test/index.test.js
```diff
@@ -0,0 +1,17 @@
+const browserslist = require( 'browserslist' );
```
gziolo: Yes, whole file should follow WordPress spacing rules.

packages/browserslist-config/test/index.test.js
ntwb: I'm going to skip this change, for now, we need to get ESLint setup

The blocker to this is getting the changes from https://make.wordpress.org/core/2017/09/07/proposal-for-js-standards-revision-removing-array-function-whitespace-exceptions/ merged.

In me hacking in ESLint to run over this PR both the following return no ESLint errors:
• `const result = browserslist([ 'extends @wordpress/browserslist-config' ]);`
• `const result = browserslist(['extends @wordpress/browserslist-config']);`

And this does:
• `const result = browserslist( ['extends @wordpress/browserslist-config'] );`

I'm not sure why `space-in-parens` ESLint rule is throwing that, but I'll get it fixed over the weekend, or early next week, a public holiday here today so I'm going to head out 🏖

packages/browserslist-config/package.json
ntwb: As `browserslist` is only used in the Jest tests it only requires to be included in `devDependencies`, that said, here's why it is included:

I added it to `peerDependencies` in https://github.com/ntwb/browserslist-config-wordpress/commit/8c6f1b10d9a0c0d43fb2ec95836ac9ae5ac98132 because of `eslint-plugin-node`'s [`no-unpublished-require`](https://github.com/mysticatea/eslint-plugin-node/blob/b0618d4a21a5ce5303f58e125f28705d4d3f7b47/docs/rules/no-unpublished-require.md) rule. 

I opened an issue a while ago [here](https://github.com/mysticatea/eslint-plugin-node/issues/47) asking that `devDependencies` modules be ignored from these rules and the end result was not what I had hoped for, hence the addition of `browserslist` to `peerDependencies` works as a workaround.


These are the two ESLint errors thrown when `browserslist` is not included in `peerDependencies`
```
  1:31  error  "browserslist" is not found                       node/no-missing-require
  1:31  error  "browserslist" is not published                   node/no-unpublished-require
```

I'm going to remove it from `peerDependencies` and add an `.eslintrc.json` in the `test` folder to ignore those two `eslint-plugin-node/recommended` rules as suggested [here](https://github.com/mysticatea/eslint-plugin-node/issues/47#issuecomment-335606914) which I think is not quite as hacky...

packages/browserslist-config/CHANGELOG.md
ntwb: Thanks @atimmer 👍 Fixed in https://github.com/WordPress/packages/pull/69/commits/398fae0f0456da6207ffff0849c508b2382135dc

packages/browserslist-config/package.json
ntwb: I removed it https://github.com/WordPress/packages/pull/69/commits/0dfaed0b902ff6cc8a89e0ca78fcf06b272b7f87, will add the `. eslintrc.json` when adding ESLint  in, I'm tempted to turn both those rules off by default config now.

PR #70, opened 2018-01-25
📦 NEW: Update Lerna to 2.8.0

ahmadawais:
Lerna 2.8.0 is a much better version and fixes a lot of problems that I personally faced. While this repo is using `2.0.0` — let's update this?


lerna.json
```diff
@@ -1,7 +1,5 @@
 {
-	"lerna": "2.0.0",
-	"packages": [
-		"packages/*"
-	],
+	"lerna": "2.8.0",
```
ntwb: This remains at `2.0.0`, this is a Lerna specific version, not the npm module packaged version.

ntwb:
This isn't required, it's covered by semantic versioning, `^2.5.1` 

You can see this at https://semver.npmjs.com/ (Screenshot https://cl.ly/3e3m2w3G0a01)

aduth:
@ntwb With `package-lock.json`, isn't it the case that we'd always be installing whichever version is defined there?

https://github.com/WordPress/packages/blob/5d23873ff5955983f50b70f2a0efc76cd851477a/package-lock.json#L1956-L1957

ahmadawais:
@ntwb That's not true!

`lerna.json` needs to be updated as well — otherwise there will be version mismatch issues and I strongly recommend against having a global installation of lerna. 

`lerna` should always be run via npm scripts to make sure we are using the lcoal package to keep everyone in sync.

lerna.json
ahmadawais: That should be updated to update lerna. Otherwise, there will be a version mismatch.

ntwb:
My bad, apologies y'all. I forgot that lockfiles break semantic versioning, let's get rid of all the `package-lock.json` files and let semantic versioning do its thing.

Right you are also @ahmadawais, sorry, again, I've got a bit to learn about Lerna 😏 
> https://github.com/lerna/lerna#lernajson
> `lerna:` the current version of Lerna being used.

p.s. I've been meaning write a post on _"lockfiles break semantic versioning"_ for a while, here's my post on why I don't believe lockfiles should be used in packages

> https://netweb.blog/2018/01/26/lockfiles-breaks-semantic-versioning/
-----

Not sure why I can't reopen this pull request, I've not seen the `The master branch was force-pushed or recreated` message before when hovering over the "Reopen pull request" button /shrug

@ahmadawais Can you create a fresh PR please and I'll merge it straight away

PR #72, opened 2018-01-26
📦 NEW: Update Lerna to 2.8.0

ahmadawais:
Lerna 2.8.0 is a much better version and fixes a lot of problems that I personally faced. While this repo is using 2.0.0 — let's update this!

 — same PR as #70 @ntwb 👍

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,7 +1,7 @@
 {
-	"lerna": "2.0.0",
-	"packages": [
-		"packages/*"
-	],
-	"version": "independent"
+  "lerna": "2.8.0",
+  "packages": [
+    "packages/*"
+  ],
+  "version": "independent"
 }
--- a/lerna.json
+++ b/lerna.json
@@ -1,7 +1,7 @@
 {
-  "lerna": "2.8.0",
-  "packages": [
-    "packages/*"
-  ],
-  "version": "independent"
+	"lerna": "2.8.0",
+	"packages": [
+		"packages/*"
+	],
+	"version": "independent"
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -108,6 +108,12 @@
 			"integrity": "sha1-104bh+ev/A24qttwIfP+SBAasjQ=",
 			"dev": true
 		},
+		"async": {
+			"version": "1.5.2",
+			"resolved": "https://registry.npmjs.org/async/-/async-1.5.2.tgz",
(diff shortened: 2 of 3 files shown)


lerna.json
```diff
@@ -1,7 +1,7 @@
 {
```
gziolo: @ntwb, we really need to add Eslint to make sure we don't need to deal with all that code formatting issues like tabs vs spaces.

lerna.json
```diff
@@ -1,7 +1,7 @@
...
-	"lerna": "2.0.0",
-	"packages": [
-		"packages/*"
-	],
-	"version": "independent"
+  "lerna": "2.8.0",
```
gziolo: They use 2.0.0 in the official docs, it seems strange to have to update the version in both places. Shouldn't we reach out Lerna team and clarify that?

lerna.json
ahmadawais: ESLint added at any stage as in after this PR would easily address these. I'll make a separate PR for that today :)

lerna.json
ahmadawais: Nops! They recommend 2.x.x version.

ahmadawais: How did you initialize the lerna repo? You should run something like

gziolo: Actually lerna config was already here, that's why I had no idea how it should work :)

lerna.json
gziolo: Yes, that's why I pinged Stephen.

ahmadawais: I have a working copy inside create-guten-block if you'd like to take a look at it?

lerna.json
ahmadawais: Glad I could help. Let's merge this one :) 👋

lerna.json
gziolo: 👌

ntwb: Yup, I'll take care of ESLint as stated in the other PR in the coming days

ntwb:
Thanks @ahmadawais 👍

ahmadawais:
Sure thing, @ntwb — would be contributing more. This is an exciting project for me.

ntwb:
> They use 2.0.0 in the official docs, it seems strange to have to update the version in both places. Shouldn't we reach out Lerna team and clarify that?

@gziolo I've created https://github.com/lerna/lerna/issues/1237 asking to help clarify a few things for this

gziolo:
@ntwb, thanks for doing that, it would be nice to have it simplified 👍

PR #77, opened 2018-02-02
Build: Symlink all child packages instead of using versions published to npm

gziolo:
When working on #74 I encountered a few issues with the way all child packages are managed and distributed. To publish `@wordpress/jest-presets-default` and make it work with the repository I had to open 2 follow-up PRs: #75 and #76 and re-publish packages to make everything work properly. This PR tries to solve this kind of issues for the good.

This is what I came up with after some discussions in https://github.com/lerna/lerna/issues/1174. I tried using `npm link`, but it created symlinks through the folder that stores globally installed packages. Another issue is it that it still tries to use dependencies published to npm for the child dependencies :) Let me try to explain it with the folder structure we have:

```
packages
    - jest-console
    - jest-preset-default
        - node_modules
           - jest-console (symlink)
```

When you create links to those folders using `npm link` they are properly placed inside:

```
node_modules
    - @wordpress
        - jest-console (symlink)
        - jest-preset-default (symlink)
```

However, the side effect of this command is that it installs all npm dependencies inside subfolders of each of the packages replacing symlinks created by Lerna with the regular npm packages ...

```
packages
    - jest-console
    - jest-preset-default
        - node_modules
           - jest-console
```

In effect `jest-console` is a copy of the latest version of `jest-console` published to npm.

What I propose here is manually created symlinks for things that we need to make tests pass. It's not the best experience but solves all other issues we have.

There are no plans to fix that inside Lerna:

> Probably not. I’m working on some mad science to completely remove the bootstrap command, which seems more worthwhile, tbh.

> I was talking about the bootstrap, not the linking. Currently I'm not doing anything on that regard

### Testing

* Run `npm i` on a clean repository.
* Repeat a few times `npm i` to make sure nothing breaks.
* Run `npm test` and make sure all tests pass.

I tested on macOS. I'm sure it will break on Windows :(


gziolo:
@ntwb I'm looking forward to see `package-lock.js` completely removed. Every time I change something in any of `package.json` file my diff shows a ridiculous number of lines changed ...

ntwb:
Regarding the symlinks, maybe we could use Lerna's `symlink` function or replicate similar functionality here in `/packages` so that we can include Windows compatible symlinks
* https://github.com/lerna/lerna/blob/master/src/FileSystemUtilities.js#L103-L110

Alternatively, https://www.npmjs.com/package/symlink-or-copy could be used

There's quite a few _symlink_ packages available https://www.npmjs.com/browse/keyword/symlink, I'm surprised Lerna dosn't actually use one of these packages.

Anyway, let's get this merged and explore the above in a follow up PR

gziolo:
I will try to use existing `npm` packages similar to `rimraf` which work on all platfoms in a follow-up PR 👍

PR #80, opened 2018-02-10
docs: add link to browserslist usage examples docs

ntwb:
Add a Babel usage example


packages/browserslist-config/README.md
```diff
@@ -19,3 +19,19 @@ Add this to your `package.json` file:
 	"extends @wordpress/browserslist-config"
 ]
 ```
+
+This package when imported returns an array of supported browsers, for example, configuring [babel-preset-env](https://github.com/babel/babel/tree/master/experimental/babel-preset-env):
+
+```js
+const browserslistWordPress = require("@wordpress/browserslist-config");
```
aduth: Should our examples adhere to coding standards?

- Single quote
- Spaces between parens
- Spaces inside (array) square brackets
- Tabs for indentation

aduth: Support for `.babelrc.js` was only added in Babel 7 (currently prerelease) so this could cause some confusion for developers trying to use the example.

https://github.com/babel/babel/issues/4630

aduth: Nit: Maybe a shorter variable name with qualification as prefix? `wpBrowserslist`

gziolo: We have this included in the preset which uses JS. I think this is where it comes from :)

ntwb: Thanks, my bad, this was me being lazy using the GitHub.com editor 😞 

That said, I'm going to remove this example anyway, reading further at https://github.com/ai/browserslist-example#babel _"From Babel 7.0 `@babel/preset-env` loads target browsers from the same Browserslist config."_ which is what we're doing in [@wordpress/babel-preset-default](https://github.com/WordPress/packages/blob/master/packages/babel-preset-default/index.js#L9) already. 

I've updated this PR to link to the official Browserslist examples docs, less maintanence for us and documents the above Babel v7 usage etc

gziolo:
I added `.` at the end of the sentence. It's good to go :)

PR #85, opened 2018-02-19
Hooks: Micro-optimize runHooks

aduth:
Foreseeing `runHooks` becoming a hot code path, I took to making a few small micro-optimizations, improving the handled and unhandled hook performance by approximately 42% and 172% respectively on my machine (handled meaning at least one hook handler exists). Standard benchmarking caveats apply (Node V8 engine, machine-specific, benchmarks themselves limited in real-world applicability).

**Before:**

```
handled x 5,839,917 ops/sec ±0.32% (91 runs sampled)
unhandled x 7,967,106 ops/sec ±0.63% (89 runs sampled)
```

**After:**

```
handled x 8,306,775 ops/sec ±0.46% (92 runs sampled)
unhandled x 22,606,859 ops/sec ±0.65% (90 runs sampled)
```

Optimizations include:

- Removing unnecessary code (a0e6eb6)
   - Technically, this removal would allow for `doAction( '__current' )` to attempt to find handlers for the internal property, though would fail after the second condition, since array would not have `.handlers` property. To me, it's more concerning that we're allowing for this special key to exist with special treatment in the same scope as other hook handlers.
- Avoid assignment when either unnecessary or redundant (71bd540, 9177da1)
- Performant alternatives (truthy property access on null-prototype object vs. `hasOwnProperty`, faa0ed3, [jsperf](https://jsperf.com/object-create-null-vs-hasownproperty))
   - The reason for `Object.create( null )` is to avoid issues with prototype property access:
      - `!! ( {} ).valueOf // true`
      - `!! Object.create( null ).valueOf // false`

Try yourself:

```
npm install
npm run build
node packages/hooks/benchmark
```

### DIFF
--- a/packages/hooks/src/createHooks.js
+++ b/packages/hooks/src/createHooks.js
@@ -12,8 +12,8 @@ import createDidHook from './createDidHook';
  * @return {Object} Object that contains all hooks.
  */
 function createHooks() {
-	const actions = {};
-	const filters = {};
+	const actions = Object.create( null );
+	const filters = Object.create( null );
 
 	return {
 		addAction:        createAddHook( actions ),
--- a/packages/hooks/src/createHooks.js
+++ b/packages/hooks/src/createHooks.js
@@ -14,6 +14,8 @@ import createDidHook from './createDidHook';
 function createHooks() {
 	const actions = Object.create( null );
 	const filters = Object.create( null );
+	actions.__current = [];
+	filters.__current = [];
 
 	return {
 		addAction:        createAddHook( actions ),
--- a/packages/hooks/src/createRunHook.js
+++ b/packages/hooks/src/createRunHook.js
@@ -21,16 +21,12 @@ function createRunHook( hooks, returnFirstArg ) {
 	 * @return {*}               Return value of runner, if applicable.
 	 */
 	return function runHooks( hookName, ...args ) {
-		if ( ! hooks[ hookName ] ) {
-			hooks[ hookName ] = {
-				runs: 0,
-				handlers: [],
-			};
+		let handlers;
+		if ( hooks[ hookName ] ) {
+			handlers = hooks[ hookName ].handlers;
 		}
 
-		const handlers = hooks[ hookName ].handlers;
-
-		if ( ! handlers.length ) {
+		if ( ! handlers || ! handlers.length ) {
 			return returnFirstArg
 				? args[ 0 ]
 				: undefined;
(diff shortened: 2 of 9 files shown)


packages/hooks/src/createRunHook.js
```diff
@@ -21,21 +21,12 @@ function createRunHook( hooks, returnFirstArg ) {
 	 * @return {*}               Return value of runner, if applicable.
 	 */
 	return function runHooks( hookName, ...args ) {
-
-		if ( ! validateHookName( hookName ) ) {
```
adamsilverstein: This removes all validation from hooks names when applying them.

packages/hooks/src/createHooks.js
```diff
@@ -12,8 +12,10 @@ import createDidHook from './createDidHook';
  * @return {Object} Object that contains all hooks.
  */
 function createHooks() {
-	const actions = {};
-	const filters = {};
+	const actions = Object.create( null );
```
adamsilverstein: Not clear why this is changing this here, is this performance related or required for other changes here?

packages/hooks/src/createRunHook.js
aduth: >This removes all validation from hooks names when applying them.

What's the worst case scenario here?

The assumption I'd made, noted in extended commit description of a0e6eb6, is that validation of the hooks takes place when they're added, and if no hooks exist which match the hook name (a given if it's an invalid hook name), nothing would occur.

packages/hooks/src/createHooks.js
aduth: Yes, from original comment:

>- Performant alternatives (truthy property access on null-prototype object vs. hasOwnProperty, faa0ed3, [jsperf](https://jsperf.com/object-create-null-vs-hasownproperty))
>   - The reason for Object.create( null ) is to avoid issues with prototype property access:
>      - `!! ( {} ).valueOf // true`
>      - `!! Object.create( null ).valueOf // false`

packages/hooks/src/createRunHook.js
adamsilverstein: Seems reasonable 👍

packages/hooks/src/createHooks.js
adamsilverstein: Ok, got it - I didn't see that this was required for the other changes to work. Nice!

adamsilverstein:
Looks good to me!

PR #88, opened 2018-02-22
Remove npm's `package-lock.json` files

ntwb:
> [Lockfiles for apps, but not for packages](https://github.com/sindresorhus/ama/issues/479#issuecomment-310661514)
> **Lockfiles for apps, but not for packages**. Lockfiles are great for apps where you want a controlled reproducible environment, but for packages this doesn't make much sense. The `package-lock.json` files in dependencies are ignored, so the lockfile only applies when users run `npm install` in the package repo. If you use a lockfile for packages, your local dependency tree will not match the dependency tree of users having your package as a dependency. This can potentially cause problems if one of your package dependencies breaks your package in a patch release. The lockfile will prevent you from seeing the problem locally, but it will affect your users.


lerna.json
```diff
@@ -1,5 +1,10 @@
 {
 	"lerna": "2.9.0",
+	"command": {
+		"bootstrap": {
+			"npmClientArgs": ["--no-package-lock"]
```
gziolo: Do we need it in the case where we have `.npmrc` file with the same flag?

ntwb: Not sure to be honest, but looking at [test/integration/lerna-bootstrap.test.js#L53](https://github.com/lerna/lerna/blob/ba794660e66242bff7bbd04397c5932468d6769a/test/integration/lerna-bootstrap.test.js#L53):
`test("--npm-client npm -- --no-optional extends durable npmClientArgs", async () => {` 

If Lerna is just using the `npm` command then Lerna should honor the contents of the `.npmrc` file I expect

ntwb: Removed in https://github.com/WordPress/packages/commit/56185d5964fc65c955c3b109a7251767618c0123

gziolo: It's still looks good. Seems to use `npm` as it is :)

PR #91, opened 2018-03-17
feat: add `jest-preset-node` package

ntwb:
This pull request seeks to add a new package `@wordpress/jest-preset-node`

This package differs from [@wordpress/jest-preset-default](https://github.com/WordPress/packages/tree/master/packages/jest-preset-default) in that currently the 3 [WordPress Coding Standards](https://github.com/WordPress-Coding-Standards) that would adopt this are basic Node.js packages and do not require React or Enzyme.

It is a pretty basic Jest configuration that would be shared between 3 repos and in the spirit of [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) principles the following 3 repos would replace the current code with:

```json
{
  "preset": "@wordpress/jest-preset-node"
}
```

> [WordPress-Coding-Standards/eslint-plugin-wordpress/blob/master/package.json#L79-L88](https://github.com/WordPress-Coding-Standards/eslint-plugin-wordpress/blob/master/package.json#L79-L88)
> [WordPress-Coding-Standards/eslint-config-wordpress/blob/master/package.json#L60-L69](https://github.com/WordPress-Coding-Standards/eslint-config-wordpress/blob/master/package.json#L60-L69)
> [WordPress-Coding-Standards/eslint-config-wordpress/blob/master/package.json#L60-L69](https://github.com/WordPress-Coding-Standards/eslint-config-wordpress/blob/master/package.json#L60-L69)

Edit: Renamed `wpcs` references to `node`


packages/jest-preset-wpcs/README.md
```diff
@@ -0,0 +1,35 @@
+# @wordpress/jest-preset-wpcs
```
youknowriad: Why not `jest-preset-node` to clarify the difference with the default package? `wpcs` looks like a duplicate of the `@wordpress` namespace

ntwb: Thanks @youknowriad I had originally started out using `jest-preset-node` and then switched to `jest-preset-wpcs` due to the Jest configuration I was using, I've refactored that now to be more generic and (I think) this will work.

I've now renamed the package to `jest-preset-node` 👍

aduth:
It would be nice if presets could be composed, where the default preset could be a superset of the more minimal configuration.

ntwb:
I'm going to work on adding support for `jest-preset-node` and `jest-preset-eslint` to `jest-preset-default` and close/replace this PR and #95

PR #93, opened 2018-03-18
Package: Add `@wordpress/custom-templated-path-webpack-plugin` package

aduth:
This pull request seeks to add a new package: `@wordpress/custom-templated-path-webpack-plugin`. This has been extracted from an inline class definition in [Gutenberg's `webpack.config.js`](https://github.com/WordPress/gutenberg/blob/8434cc93d04d70543f26994f8b75cac51aac995b/webpack.config.js#L81-L127), and updated for Webpack 4+ .

Refer to [included README.md](https://github.com/WordPress/packages/blob/712e5996af3b00e76b1b70d23aee0c3c42d33455/packages/custom-templated-path-webpack-plugin/README.md) for more information.

__Testing instructions:__

See: https://github.com/WordPress/gutenberg/pull/5687

### DIFF
--- a/packages/custom-templated-path-webpack-plugin/package.json
+++ b/packages/custom-templated-path-webpack-plugin/package.json
@@ -1,0 +1,33 @@
+{
+	"name": "@wordpress/custom-templated-path-webpack-plugin",
+	"version": "1.0.0",
+	"description": "Webpack plugin for creating custom path template tags",
+	"author": "WordPress",
+	"license": "GPL-2.0-or-later",
+	"keywords": [
+		"webpack",
+		"webpack-plugin"
+	],
+	"homepage": "https://github.com/WordPress/packages/tree/master/packages/custom-templated-path-webpack-plugin/",
+	"repository": {
+		"type": "git",
+		"url": "https://github.com/WordPress/packages.git"
+	},
+	"bugs": {
+		"url": "https://github.com/WordPress/packages/issues"
+	},
+	"main": "build/index.js",
+	"module": "build-module/index.js",
+	"publishConfig": {
+		"access": "public"
+	},
+	"dependencies": {
+		"escape-string-regexp": "^1.0.5"
+	},
+	"devDependencies": {
+		"webpack": "^4.1.1"
+	},
+	"peerDependencies": {
+		"webpack": "^4.0.0"
+	}
+}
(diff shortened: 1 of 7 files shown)


packages/custom-templated-path-webpack-plugin/package.json
```diff
@@ -0,0 +1,33 @@
+{
+	"name": "@wordpress/custom-templated-path-webpack-plugin",
+	"version": "1.0.0",
```
ntwb: This initial version here needs to *not* be `1.0.0`, @gziolo has used `"version": "1.0.1-0",` in the past which has worked fine though I think we should use `1.0.0-alpha`. 

This all because Lerna _bumps_ this version upon release, if it were to stay at `1.0.0` then Lerna would only be able to publish a _patch_ `1.0.1`, _minor_ `1.1.0`, or _major_ `2.0.0` release and a `1.0.0` version would never be published.

aduth: Well, that's annoying 😄 Maybe it'd be fine to publish an initial release manually, without Lerna?

packages/custom-templated-path-webpack-plugin/test/fixtures/webpack.config.js
```diff
@@ -0,0 +1,39 @@
...
+	context: __dirname,
+	entry: './entry',
+	output: {
+		filename: '[basename].js',
+		path: __dirname,
+	},
+	plugins: [
+		new CustomTemplatedPathPlugin( {
+			basename( path, data ) {
+				console.log(data)
```
gziolo: I think it wasn’t expected to be landed 😃

aduth: 😞

aduth: At least it's in a test file.

gziolo: I need to add this to jest-console 😎

gziolo:
By the way, what was the reason you avoided import/export statements? In src subfolder it’s perfectly fine to use all Babel friendly code. Build step generates 2 versions by default.

aduth:
Because it doesn't transpile to CommonJS-compatible. If using `export default`, consuming projects need to require via `require( '@wordpress/custom-templated-path-webpack-plugin' ).default` (note `.default`).

gziolo:
Right, that reminds me I sort of did the same for node based packages 😃
The only difference is that I put them in the root folder to skip Babel magic. Both work similar.
I think it isn’t the case for tests though.

PR #94, opened 2018-03-19
Add `publish:check` script to run `lerna updated`

ntwb:
This PR adds a new npm script `publish:check`, by running `npm run publish:check` Lerna will check which packages have changed (via `lerna updated`) since the last publish (via `lerna publish`).

The script also runs the _build_ scripts for additional pre-release checks
```shell
npm run publish:check
lerna info version 2.9.0
lerna info versioning independent
lerna info ignore [ '**/test/*.js' ]
lerna info Checking for updated packages...
lerna info Comparing with @wordpress/a11y@1.0.7.
lerna info Checking for prereleased packages...
lerna info result 
- @wordpress/jest-preset-default
- @wordpress/scripts
```

Note: The script also runs the _build_ scripts for additional pre-release checks though this isn't displayed in the above example output.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
 		"test:coverage": "npm run test -- --coverage",
 		"test:coverage-ci": "npm run test -- --coverage && codecov",
 		"test:watch": "npm run test -- --watch",
+		"publish:check": "npm run build && lerna updated",
 		"publish:dev": "npm run build && lerna publish --npm-tag next",
 		"publish:prod": "npm run build && lerna publish"
 	}


aduth:
Can you explain why the build pre-step is necessary? How does Lerna determine when updates are needed? My impression was that it was looking at commits since the last tag which affect files in a given directory; where a build would not alter the output of `lerna updated`.

gziolo:
Yes, interesting question from @aduth. Apart from that, the command `lerna updated` is pretty useful 👍

ntwb:
> Can you explain why the build pre-step is necessary?

You are correct @aduth in that it is not necessary

In creating the PR I copied and pasted the `publish:prod` line and in making the patch I asked myself the very same question, I decided to leave it there as a bit of a sanity check, if you are running `lerna updated` then this would typically be because you're starting to perform checks of what would be required for a new release and this extra check if it should fail will prompt you that the repo is not in a pristine state and should be investigated further before continuing on on the road to publishing any package updates

PR #96, opened 2018-03-22
WordPress i18n package: Initial commit

youknowriad:
This pull request seeks to add a new package `@wordpress/i18n` package extracted from Gutenberg.

 - It includes the JavaScript library used to translate the strings
 - A babel plugin and node script needed to make sure your plugin is i18n ready in the WordPress.org repository (these two scripts could be removed later if the repository implements JavaScript strings discovery).


adamsilverstein:
@youknowriad thank you for opening this PR! I see codedev complaining about reduced tests, can we add more tests to increase coverage?

youknowriad:
@adamsilverstein I think it's pretty well tested, what's not tested is the `pot-2-php` script which is just a temporary script to generate a dummy php file (for discovery in .org). I think this file is hurting the coverage but I don't think it's worth testing.

youknowriad:
I added unit tests, the coverage is still decreasing because of `pot-to-php` script but as I said, this one is not worth testing.

ocean90:
The usage of `pot-to-php` should not be promoted like this, because using one file for all strings with `get_translations_for_domain()` is not really a solution. I suggest to not include it in this package. 

String extraction to translate.w.org can be added quite easily once we have identified how and which translations should get extracted. See https://core.trac.wordpress.org/ticket/20491#comment:82.

youknowriad:
> String extraction to translate.w.org can be added quite easily once we have identified how and which translations should get extracted. See https://core.trac.wordpress.org/ticket/20491#comment:82.

That would be great, I'm not aware of the work being done to do so but I believe this PR identifies properly what should be extracted (all the APIs documented in the README). Who can help with this?

I'd be happy to remove the `pot-to-php` once done but right now we need a solution for it. I believe the `babel` plugin is useless as well once the extraction lands.

ocean90:
> but right now we need a solution for it. 

With "we" you mean Gutenberg? Don't see why that script can't remain there as a standalone script. Since the package should only be published once core and w.org fully supports this proposal for the JS i18n API I don't see the need for the script.

> I believe the babel plugin is useless as well once the extraction lands.

The babel plugin doesn't seem useless since that's what w.org is probably going to use for the extraction. Same for authors that don't host their plugins/themes on w.org.

The current extraction for PHP can be found here: https://meta.trac.wordpress.org/browser/sites/trunk/wordpress.org/public_html/wp-content/plugins/plugin-directory/cli/i18n/class-code-import.php
So for the JS strings we need something that can be run similar to this. 


> I believe this PR identifies properly what should be extracted

How does it handle a JS file with multiple text domains? Can a fixed text domain be passed to the script? Can it handle `const __ = (text, domain = 'gutenberg') => {...}` as proposed in #4147? Does it work on build files?

youknowriad:
> With "we" you mean Gutenberg?

I was think "we" as Plugin authors, Core developpers, if we land this package as a generic package we need a way to make these strings translatable. If not, the package is useless.

> How does it handle a JS file with multiple text domains? Can a fixed text domain be passed to the script? Can it handle const __ = (text, domain = 'gutenberg') => {...} as proposed in #4147? 

Yes, it does the domain is the last arg for all these functions.

> Does it work on build files,

That's a good point and I don't really know. Right now we use it on unminified files only. The problem with minification... is that we don't control the tool doing it, it'd be hard to come up with a solution that works consistently across tools.  Thoughts @aduth

aduth:
The Babel plugin itself doesn't make distinguish on domain. It will extract the strings, regardless of the domain passed (if any), but there is only a single output file, so a source file with two separate domains would still be processed into a single `.pot` file.

Minification could be an issue with "mangling", where minifiers choose a shorter auto-generated variable name to use in place of the original name. Since the plugin relies on the names of the functions (`__`, etc), these strings would not be sourced. At least for UglifyJS, there are options to exempt certain reserved variables to avoid minification:

https://github.com/mishoo/UglifyJS2#cli-mangle-options

I expect this may be most problematic for trying to extract strings from the plugin repository. During a plugin's own build process, this is unlikely to occur, since the Babel plugin would apply its transforms before minification is applied.

packages/i18n/tools/pot-to-php.js
```diff
@@ -0,0 +1,135 @@
+#!/usr/bin/env node
```
aduth: Did you run this file through some formatter? It's got all sorts of styling issues which aren't present in the same file from the Gutenberg repository (spaces instead of tabs, double quotes instead of single quotes).

packages/babel-plugin-i18n/README.md
```diff
@@ -0,0 +1,18 @@
+i18n
```
aduth: Re: Naming. I think the title here should reflect the fact that it's a Babel plugin. Maybe at least "i18n Babel Plugin".

But aside from it being scoped to WordPress, and WordPress using gettext for i18n, I feel a bit odd calling this a "i18n Babel Plugin", since gettext is just one method of internationalizing.

The equivalent PHP tool is called `makepot.php`, so maybe `babel-plugin-makepot` if we want consistency (whether or not the name on its own is great).

Otherwise some variations of words in `babel-plugin-extract-gettext-strings` ? `-extract-strings`? `-extract-gettext` ? `-gettext-strings` ?

packages/i18n/tools/pot-to-php.js
youknowriad: Probably my default prettier setup :P I'll fix it

packages/babel-plugin-i18n/README.md
youknowriad: `babel-plugin-makepot` sounds good to me

youknowriad:
Feedback addressed, this should be in a better shape.

packages/babel-plugin-makepot/package.json
```diff
@@ -0,0 +1,32 @@
...
+	],
+	"homepage": "https://github.com/WordPress/packages/tree/master/packages/babel-plugin-makepot/README.md",
+	"repository": {
+		"type": "git",
+		"url": "https://github.com/WordPress/packages.git"
+	},
+	"bugs": {
+		"url": "https://github.com/WordPress/packages/issues"
+	},
+	"main": "src/index.js",
```
aduth: The root build configuration is such that a build distributable will be created for this, even if the `src/index.js` doesn't require transpilation. We should consider one of:

- Target `build/index.js` as `main`, even if not strictly necessary, merely for consistency with other packages
- Introduce a pattern for excluding particular projects from the Babel build step

packages/i18n/README.md
```diff
@@ -0,0 +1,76 @@
...
+// 4 hats
+```
+
+Note that you will not need to specify [domain](https://codex.wordpress.org/I18n_for_WordPress_Developers#Text_Domains) for the strings.
+
+## Build
+
+You can use the [WordPress i18n babel plugin](../babel-plugin-makepot/README.md) to generate a `.pot` file containing all your localized strings.
+
+The package also includes a `pot-to-php` script used to generate a php files containing the messages listed in a `.pot` file. This is usefull to trick WordPress.org translation strings discovery since at the moment, WordPress.org is not capable of parsing strings directly from JavaScript files.
```
aduth: Typo: "usefull" -> "useful"

packages/i18n/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+```sh
+node tools/pot-to-php languages/myplugin.pot languages/myplugin-translations.php text-domain
```
aduth: I think we want to make this a proper `bin` entry in `package.json`:

https://docs.npmjs.com/files/package.json#bin

That way, a developer could execute via `npx pot-to-php`.

As-is, the code snippet isn't very usable (would need to at least be prefixed with `./node_modules/@wordpress/i18n`).

packages/i18n/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+See: https://developer.wordpress.org/reference/functions/_nx/
+
+`sprintf( format: string, ...args: mixed[] ): string`
+
+Returns a formatted string.
+
+See: http://www.diveintojavascript.com/projects/javascript-sprintf
+
+`setLocaleData( data: Object, domain: string )`
```
aduth: Thanks for fixing the documentation I keep neglecting to fix 😄

packages/babel-plugin-makepot/package.json
youknowriad: I'm using the build folders right now, let's add this option separately because it seems useful for other packages as well.

youknowriad:
From yesterday's chat, there's no blocking here. So moving forward.

Issue #104, opened 2018-04-09
Support an optional namespace parameter for hasAction/hasFilter (matching core)


adamsilverstein:
See: https://github.com/WordPress/packages/issues/78#issuecomment-379860775

ntwb:
Issue moved to [WordPress/gutenberg #7803](https://github.com/WordPress/gutenberg/issues/7803) via [**ZenHub**](https://www.zenhub.com/)

PR #106, opened 2018-04-10
TRY: Support an optional namespace parameter for hasAction/hasFilter

adamsilverstein:
* Add tests
* Update docblock

Fixes https://github.com/WordPress/packages/issues/104

### DIFF
--- a/packages/hooks/src/createHasHook.js
+++ b/packages/hooks/src/createHasHook.js
@@ -20,7 +20,7 @@ function createHasHook( hooks ) {
 		// Use the namespace if provided.
 		if ( 'undefined' !== typeof namespace ){
 			return hookName in hooks &&
-				hooks[ hookName ].handlers.some( hook => hook['namespace'] === namespace );
+				hooks[ hookName ].handlers.some( hook => hook.namespace === namespace );
 		}
 
 		return hookName in hooks;
--- a/packages/hooks/src/createHasHook.js
+++ b/packages/hooks/src/createHasHook.js
@@ -9,7 +9,7 @@
  */
 function createHasHook( hooks ) {
 	/**
-	 * Returns how many handlers are attached for the given hook.
+	 * Returns whether any handlers are attached for the given hookName and optional namespace.
 	 *
 	 * @param  {string}  hookName  The name of the hook to check for.
 	 * @param  {string}  namespace Optional. The unique namespace identifying the callback in the form `vendor/plugin/function`.
--- a/packages/hooks/src/createHasHook.js
+++ b/packages/hooks/src/createHasHook.js
@@ -11,8 +11,8 @@ function createHasHook( hooks ) {
 	/**
 	 * Returns whether any handlers are attached for the given hookName and optional namespace.
 	 *
-	 * @param  {string}  hookName  The name of the hook to check for.
-	 * @param  {string}  namespace Optional. The unique namespace identifying the callback in the form `vendor/plugin/function`.
+	 * @param {string} hookName  The name of the hook to check for.
+	 * @param {string} namespace Optional. The unique namespace identifying the callback in the form `vendor/plugin/function`.
 	 *
 	 * @return {boolean} Whether there are handlers that are attached to the given hook.
 	 */
--- a/packages/hooks/src/createHasHook.js
+++ b/packages/hooks/src/createHasHook.js
@@ -20,7 +20,6 @@ function createHasHook( hooks ) {
 		// Use the namespace if provided.
 		if ( 'undefined' !== typeof namespace ){
 			return hookName in hooks &&
-				hooks[ hookName ].handlers &&
 				hooks[ hookName ].handlers.some( hook => hook['namespace'] === namespace );
 		}
 
--- a/packages/hooks/src/createHasHook.js
+++ b/packages/hooks/src/createHasHook.js
@@ -5,17 +5,24 @@
  * @param  {Object}   hooks Stored hooks, keyed by hook name.
  *
  * @return {Function}       Function that returns whether any handlers are
- *                          attached to a particular hook.
+ *                          attached to a particular hook and optional namespace.
  */
 function createHasHook( hooks ) {
 	/**
 	 * Returns how many handlers are attached for the given hook.
 	 *
-	 * @param  {string}  hookName The name of the hook to check for.
+	 * @param  {string}  hookName  The name of the hook to check for.
+	 * @param  {string}  namespace Optional. The unique namespace identifying the callback in the form `vendor/plugin/function`.
 	 *
 	 * @return {boolean} Whether there are handlers that are attached to the given hook.
 	 */
-	return function hasHook( hookName ) {
+	return function hasHook( hookName, namespace ) {
+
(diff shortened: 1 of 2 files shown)


packages/hooks/src/createHasHook.js
```diff
@@ -5,17 +5,25 @@
...
+ *                          attached to a particular hook and optional namespace.
  */
 function createHasHook( hooks ) {
 	/**
-	 * Returns how many handlers are attached for the given hook.
+	 * Returns whether any handlers are attached for the given hookName and optional namespace.
 	 *
-	 * @param  {string}  hookName The name of the hook to check for.
+	 * @param  {string}  hookName  The name of the hook to check for.
+	 * @param  {string}  namespace Optional. The unique namespace identifying the callback in the form `vendor/plugin/function`.
```
ntwb: Single spaces between `@param` and `{string}` and `name`

```js
	 * @param {string} hookName  The name of the hook to check for.
	 * @param {string} namespace Optional. The unique na
```

ntwb (changes requested):

packages/hooks/src/createHasHook.js
adamsilverstein: thanks, corrected.

packages/hooks/src/createHasHook.js
```diff
@@ -5,17 +5,25 @@
...
 	 *
 	 * @return {boolean} Whether there are handlers that are attached to the given hook.
 	 */
-	return function hasHook( hookName ) {
+	return function hasHook( hookName, namespace ) {
+		// Use the namespace if provided.
+		if ( 'undefined' !== typeof namespace ){
+			return hookName in hooks &&
+				hooks[ hookName ].handlers &&
+				hooks[ hookName ].handlers.map( hook => hook['namespace'] ).includes( namespace );
```
aduth: `Array#includes` is not polyfilled, and is not supported in older browsers:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes#Browser_compatibility

Further, there's unnecessary work here to iterate the entire array of handlers, when `Array#some` serves this purpose to terminate iteration once the first match is encountered:

```js
hooks[ hookName ].handlers.some( hook => hook['namespace'] === namespace );
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

packages/hooks/src/createHasHook.js
```diff
@@ -5,17 +5,25 @@
...
+	 * @param {string} namespace Optional. The unique namespace identifying the callback in the form `vendor/plugin/function`.
 	 *
 	 * @return {boolean} Whether there are handlers that are attached to the given hook.
 	 */
-	return function hasHook( hookName ) {
+	return function hasHook( hookName, namespace ) {
+		// Use the namespace if provided.
+		if ( 'undefined' !== typeof namespace ){
+			return hookName in hooks &&
+				hooks[ hookName ].handlers &&
```
aduth: When will this not be truthy?

packages/hooks/src/createHasHook.js
adamsilverstein: Ah, right - thanks for catching that. Will fix.

packages/hooks/src/createHasHook.js
aduth: >generally prefer the defensive approach of always checking before using because otherwise everything breaks.

Well, there's defensive, but there's also having a solid understanding of the shapes of data we expect to occur.

adamsilverstein: True :)  I'll take another look - if handlers is always going to be truthy, i will remove the check.

packages/hooks/src/createHasHook.js
adamsilverstein: updated

packages/hooks/src/createHasHook.js
adamsilverstein: removed extraneous check

kadamwhite (commented):
The implementation looks straightforward, I’d just question whether we’re pushing for this just to match core and fill a perceived need or whether we’ll actually find this useful in practice.

packages/hooks/src/createHasHook.js
```diff
@@ -5,17 +5,24 @@
...
+	 * @param {string} namespace Optional. The unique namespace identifying the callback in the form `vendor/plugin/function`.
 	 *
 	 * @return {boolean} Whether there are handlers that are attached to the given hook.
 	 */
-	return function hasHook( hookName ) {
+	return function hasHook( hookName, namespace ) {
+		// Use the namespace if provided.
+		if ( 'undefined' !== typeof namespace ){
+			return hookName in hooks &&
+				hooks[ hookName ].handlers.some( hook => hook.namespace === namespace );
```
aduth: Minor: This style of arrow function will fail ESLint when merged to Gutenberg.

https://github.com/WordPress/gutenberg/blob/ef5cfe077cb54e2b74881aeb7407b9ed15ebce85/eslint/config.js#L32
https://eslint.org/docs/rules/arrow-parens

packages/hooks/src/createHasHook.js
```diff
@@ -5,17 +5,24 @@
...
-	 * @param  {string}  hookName The name of the hook to check for.
+	 * @param {string} hookName  The name of the hook to check for.
+	 * @param {string} namespace Optional. The unique namespace identifying the callback in the form `vendor/plugin/function`.
 	 *
 	 * @return {boolean} Whether there are handlers that are attached to the given hook.
 	 */
-	return function hasHook( hookName ) {
+	return function hasHook( hookName, namespace ) {
+		// Use the namespace if provided.
+		if ( 'undefined' !== typeof namespace ){
```
aduth: Style: Space before opening curly brace.

packages/hooks/src/createHasHook.js
```diff
@@ -5,17 +5,24 @@
...
+			return hookName in hooks &&
+				hooks[ hookName ].handlers.some( hook => hook.namespace === namespace );
+		}
+
 		return hookName in hooks;
```
aduth: Minor: Regardless of whether `namespace` is provided, we have this same condition apply, so rather than duplicate, we could move it as a primary check of the function:

```js
if ( ! ( hookName in hooks ) ) {
	return false;
}

if ( 'undefined' === typeof namespace ) {
	return true;
}

return hooks[ hookName ].handlers.some( ( hook ) => {
	return hook.namespace === namespace;
} );
```

aduth (approved):
This looks good to me.

I think improved parity with equivalent core functions is enough justification to want this to land.

Unless it's urgent, I might suggest waiting 'til after https://github.com/WordPress/gutenberg/pull/7556 to avoid complications with including the new history there.

adamsilverstein:
@aduth thanks for the feedback, I'll work on these soon

ntwb:
Issue moved to [WordPress/gutenberg #7819](https://github.com/WordPress/gutenberg/issues/7819) via [**ZenHub**](https://www.zenhub.com/)

PR #108, opened 2018-04-11
TRY: Make namespace optional and combine with hookName using dot notation

adamsilverstein:
This PR seeks to remove the namespace parameter and instead enable support for an optional namespace added to the hookName using dot notation, in the form `hookName.vendor/plugin/function`. Fixes https://github.com/WordPress/packages/issues/107.

Includes updated docs: https://github.com/WordPress/packages/blob/fda1b56a606afccf40b86867f80710f6895559bb/packages/hooks/README.md

* after this change, periods are no longer allowed in the hook name - instead they are used to separate the hook name from the namespaced function name.
* when a namespace is omitted from a call to removeFilters/Actions, all callbacks on that hook name are removed.
* refactored all tests

### DIFF
--- a/packages/hooks/README.md
+++ b/packages/hooks/README.md
@@ -18,21 +18,52 @@ A lightweight & efficient filter and action manager.
 
 ### API Usage
 
+When adding a hooked action or filter, best practice is to add a namespace in the form `hookName.vendor/plugin/functionName`. This enables checking for and removing the hooked callback.
+
+// Create a hooks instance.
 * `createHooks()`
-* `addAction( 'hookName', 'functionName', callback, priority )`
-* `addFilter( 'hookName', 'functionName', callback, priority )`
-* `removeAction( 'hookName', 'functionName' )`
-* `removeFilter( 'hookName', 'functionName' )`
+
+// Add an action callback, with an optional namespace.
+* `addAction( 'hookName(.namespace)', callback, priority )` 
+
+// Add a filter callback, with an optional namespace.
+* `addFilter( 'hookName(.namespace)', callback, priority )` 
+
+// Remove all callbacks hooked on hookName.namespace. Removes all action callbacks hooked on hookName when namespace is omitted.
+* `removeAction( 'hookName(.namespace)', 'functionName' )` 
+
+// Remove all callbacks hooked on hookName.namespace. Removes all filter callbacks hooked on hookName when namespace is omitted.
+* `removeFilter( 'hookName(.namespace)', 'functionName' )` 
+
+// Remove all action callbacks hooked on hookName.
 * `removeAllActions( 'hookName' )`
+
+// Remove all filter callbacks hooked on hookName.
 * `removeAllFilters( 'hookName' )`
+
+// Invoke all action callbacks hooked on hookName, optionally passing arguments.
 * `doAction( 'hookName', arg1, arg2, moreArgs, finalArg )`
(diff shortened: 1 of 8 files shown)


gziolo:
@adamsilverstein, thanks for opening this PR. It seems like the proposed approach is the best possible moving forward. I have only one concern, which is a blocker in my opinion. This won't work with Gutenberg because of the way we name our hooks. See: https://wordpress.org/gutenberg/handbook/extensibility/extending-blocks/.

Some examples:
- `blocks.registerBlockType`
- `blocks.BlockEdit`
- `blocks.getSaveElement`
- `blocks.getSaveContent.extraProps`

It won't work properly with what you proposed and I'm afraid it's too late to change those names inside Gutenberg. Can we seek for a better naming convention that would work as you proposed, but would also ensure that Gutenberg hooks will work as before?

Some alternatives are:
* `blocks.getSaveContent.extraProps/core/generated-class-name/save-props`
* `blocks.getSaveContent.extraProps#core/generated-class-name/save-props`

I think we have 2 different styles because there were 2 different validation method for both params. We can update 2nd part to match whatever we agree upon in here.

packages/hooks/README.md
```diff
@@ -18,21 +18,52 @@ A lightweight & efficient filter and action manager.
 
 ### API Usage
 
+When adding a hooked action or filter, best practice is to add a namespace in the form `hookName.vendor/plugin/functionName`. This enables checking for and removing the hooked callback.
+
+// Create a hooks instance.
```
gziolo: It's not wrapped in code so we should either do it or remove preceding `//` :)

packages/hooks/src/validateHookName.js
```diff
@@ -19,8 +19,8 @@ function validateHookName( hookName ) {
 		return false;
 	}
 
-	if ( ! /^[a-zA-Z][a-zA-Z0-9_.-]*$/.test( hookName ) ) {
-		console.error( 'The hook name can only contain numbers, letters, dashes, periods and underscores.' );
+	if ( ! /^[a-zA-Z][a-zA-Z0-9_-]*$/.test( hookName ) ) {
+		console.error( 'The hook name can only contain numbers, letters, dashes and underscores.' );
```
gziolo: Gutenberg and its hooks depend on periods :(

gziolo (changes requested):

herregroen:
I quite like this proposal. Feels very natural and fitting for Javascript.

It could be worth exploring if adding console warnings when namespaces are omitted is something we want to encourage good standards, perhaps as an option that can be passed to `createHooks` so projects that expect multiple plugins can enable while projects that have no need for it can disable it.

Regarding the above. While periods would be preferred in my opinion due to the consistency with jQuery we could of course replace them with say a `#` or some other character.

adamsilverstein:
I took a slightly different approach in https://github.com/WordPress/packages/pull/113 by making the namespace parameter optional, but not moving it into the hook parameter. This will give us better back compat for existing use cases, and work correctly with Gutenberg.

For plugins, the idea would be to have registerPlugin pass the plugin namespace to created hooks.

```
registerPlugin( 'my-namespace', {
...
hooks: [ 
  'blocks.BlockEdit': function() { ... },
etc...
]
```

ntwb:
Issue moved to [WordPress/gutenberg #7820](https://github.com/WordPress/gutenberg/issues/7820) via [**ZenHub**](https://www.zenhub.com/)

PR #110, opened 2018-04-20
Add package: @wordpress/is-shallow-equal

aduth:
Related: https://github.com/WordPress/gutenberg/pull/6313
Related: https://github.com/WordPress/gutenberg/pull/5616

This pull request seeks to add a new package: `@wordpress/is-shallow-equal`. In my pursuit of the perfect shallow equality utility, I've come to the conclusion that for various reasons, none of the existing solutions are quite perfect. Since a function like `isShallowEqual` is often used for performance-critical code, anything less than perfection was not adequate. I've enumerated all of these reasons in the included README.md. To the point of performance specifically, contrasted with `shallowequal` as used in Gutenberg, there is shown to be a ~40% improvement.

[View README](https://github.com/WordPress/packages/blob/703d20e27f986dedbf02b36d6bdb72397e28db1e/packages/is-shallow-equal/README.md)

__Testing instructions:__

Verify tests pass:

```
npm test
```


packages/is-shallow-equal/README.md
```diff
@@ -0,0 +1,104 @@
...
+1. …consider complex-but-referentially-equal members values as equal.
+   - Sorry, [`is-equal-shallow`](https://www.npmjs.com/package/is-equal-shallow).
+2. …offer a single function through which to interface, regardless of value type.
+   - Sorry, [`shallow-equal`](https://www.npmjs.com/package/shallow-equal).
+3. …be barebones; only providing the basic functionality of shallow equality.
+   - Sorry, [`shallow-equals`](https://www.npmjs.com/package/shallow-equals).
+4. …be intended for use in non-Facebook projects.
+   - Sorry, [`fbjs/lib/shallowEqual`](https://www.npmjs.com/package/fbjs).
+5. …be the fastest implementation.
+   - Sorry, _every other solution_.
```
youknowriad: ❤️

packages/is-shallow-equal/index.js
```diff
@@ -0,0 +1,90 @@
+var isArray = Array.isArray,
+	keys = Object.keys;
```
youknowriad: Are you using `var` for performance reasons? Thinking these may create ESlint errors if we add strict rules (const, let).

packages/is-shallow-equal/index.js
```diff
@@ -0,0 +1,90 @@
...
+ * @return {Boolean} Whether the two objects are shallow equal.
+ */
+function isShallowEqualObjects( a, b ) {
+	if ( a === b ) {
+		return true;
+	}
+
+	var aKeys = keys( a ),
+		bKeys = keys( b ),
+		length = aKeys.length;
```
youknowriad: Can't we use separate `var` for each one?

packages/is-shallow-equal/index.js
```diff
@@ -0,0 +1,90 @@
...
+ *
+ * @return {Boolean} Whether the two objects are shallow equal.
+ */
+function isShallowEqualObjects( a, b ) {
+	if ( a === b ) {
+		return true;
+	}
+
+	var aKeys = keys( a ),
+		bKeys = keys( b ),
```
youknowriad: I wonder if a `for in` loop could be better than relying on keys performance wise?

youknowriad: I guess this one is better if the keys length is different?

packages/is-shallow-equal/index.js
aduth: >Are you using `var` for performance reasons? Thinking these may create ESlint errors if we add strict rules (const, let).

I'm using `var` because I'm effectively opting out of any transpilation. I think if we were to introduce such rules, we should apply them to each package's `src` folder, not necessarily a raw implementation as is done here.

packages/is-shallow-equal/index.js
aduth: `Object.keys` is actually quite fast. Faster than `for in`: https://jsperf.com/object-keys-iteration/20

And yes, there's a shortcutting opportunity here that we can assume the two values are different if they have keys of unequal length. Also not just for shortcutting, but verifying in case the second object has more keys than the first that we don't miss out on considering those (`a = { foo: 1 }, b = { foo: 1, bar: 2 }`)

packages/is-shallow-equal/index.js
aduth: >Can't we use separate `var` for each one?

We can, but my obsessive-yet-likely-often-misled interpretations of benchmark results had me do very specific code style 😄

youknowriad: But in reality, these are not that important because in general these get mangled, minified in the app bundles.

aduth: Fair point, at least for web bundles. There's still application in a Node script, which would use this code as written.

I should note that I'm not even certain this was the reason for the particular style in this case, just a general note that common style best practices which are encouraged more for maintainability's sake were deemphasized here. The file would probably need a fair few number of `eslint-disable`.

packages/is-shallow-equal/README.md
ntwb: 😂

PR #112, opened 2018-04-23
Add prependHTTP() to @wordpress/url

pento:
In the Classic Editor, typing a domain in the link UI (without the `http://`) will automatically prepend the `http://` when you save the link.

Splitting this bit of functionality out into a seperate little utility function would be useful for implementing [the same behaviour in Gutenberg](https://github.com/WordPress/gutenberg/issues/5763).

@aduth noted that there's [an NPM package](https://www.npmjs.com/package/prepend-http) that provides similar functionality, but I'm not wild about importing a third party package for such a small function.


packages/url/README.md
```diff
@@ -13,10 +13,13 @@ npm install @wordpress/url --save
 ## Usage
 
 ```JS
-import { addQueryArgs } from '@wordpress/url';
+import { addQueryArgs, prependHTTP } from '@wordpress/url';
 
 // Appends arguments to the query string of a given url
 const newUrl = addQueryArgs( 'https://google.com', { q: 'test' } ); // https://google.com/?q=test
+
+// Prepends 'http://' to URLs that are probably mean to have them
```
gziolo: Out of curiosity, why not `prependHTTPS`? :)

pento: I thought about it. For a simple function, `http` is the only viable option: sites that have `https` will generally redirect from `http` to `https`. Sites that don't have `https` will never redirect from `https` to `http`.

So, the next option is to make it more complex: perhaps try doing an XHR `HEAD` request to the URL, before deciding what protocol to prepend. I'm okay with someone else exploring that, but for now, let's just go with existing behaviour that doesn't seem to get any complaints.

rmccue: > perhaps try doing an XHR `HEAD` request to the URL, before deciding what protocol to prepend

That'll fail CORS policy, so that's not really possible either; we could proxy via the backend, but IMO `http://` is Good Enough.

pento: It should (having not actually tested it) fail with a server error, rather than a DNS or connection error. Don't need to actually talk to the server, just need to know that it's able to talk. 🙂

PR #115, opened 2018-04-25
Improve publishing process

pento:
Improve the documentation for the package release process.

- Remind people that they _must_ have 2FA enabled.
- Add instructions for using 2FA while publishing.
- Add a reminder to check for `publishConfig` settings before publishing new packages.

This also adds a check for `"publishConfig": { "access": "public" }` being set in each module's `package.json`. `npm run build` will fail (causing Travis, and publishing to fail) if it isn't set.


scripts/build.js
```diff
@@ -128,11 +129,18 @@ function buildFileFor( file, silent, environment ) {
...
+	const packageFile = packagePath + '/package.json';
 	const files = glob.sync( srcDir + '/**/*.js', { nodir: true } )
 		.filter( file => ! /\.test\.js/.test( file ) );
 
 	process.stdout.write( `${ path.basename( packagePath ) }\n` );
 
+	const config = require( packageFile );
+	if ( ! config.publishConfig || 'public' !== config.publishConfig.access ) {
+		process.stdout.write( `${ ERROR } package.json must include "publishConfig": { "access": "public" }\n` );
+		process.exitCode = 1;
```
aduth: I think with this usage, the script will continue to run, but exit with a code of `1`? vs. [`process.exit( 1 )`](https://nodejs.org/api/process.html#process_process_exit_code) which terminates immediately.

CONTRIBUTING.md
```diff
@@ -44,14 +44,28 @@ $ npm run test:coverage
...
+```json
+	"publishConfig": {
+		"access": "public"
+	}
+```
 
 ### Development release
 
-Run the following command to release a dev version of the outdated packages:
+Run the following command to release a dev version of the outdated packages, replacing "123456" with your 2FA code:
```
aduth: It may be worth making a note that since publishing can take some time, it is good to ensure that the one-time code still has a decent duration remaining before starting the process. Made more cumbersome by the fact that the developer needs to navigate through the publishing options before publishing starts.

scripts/build.js
pento: Yah, I figured it was fine to let it run to completion, so you see build errors for all modules, if there are multiple.

CONTRIBUTING.md
pento: 👍🏻

PR #117, opened 2018-04-25
Add test for arrays being completely equal

aaronjorbin:
It's a nearly useless test, but it does get is-shallow-equal to 100% test coverage
See https://codecov.io/gh/WordPress/packages/src/master/packages/is-shallow-equal/index.js#L48 for the one piece of code not currently tested.


packages/is-shallow-equal/test/index.js
```diff
@@ -104,6 +104,13 @@ describe( 'isShallowEqual', () => {
 
 		expect( isShallowEqual( a, b ) ).toBe( true );
 	} );
+
+	it( 'returns true on arrays that are a copy of each other', () => {
+		const a = [];
+		const b = a;
+
+		expect( isShallowEqual( a, b ) ).toBe( true );
+	});
```
aduth: Style: Should be space between the closing `}` and `)` (there was an exception here in the standards which was recently removed).

PR #119, opened 2018-05-06
Add `@wordpress/npm-package-json-lint-config` package

ntwb:
This PR seeks to add a new package `@wordpress/npm-package-json-lint-config`

It implements a "shared config" for the `package.json` [npm-package-json-lint](https://github.com/tclindner/npm-package-json-lint) linter

It is based on the upcoming currently unreleased v3 release of [`npm-package-json-lint`](https://github.com/tclindner/npm-package-json-lint/tree/v3)

This was briefly discussed in last weeks [#core-js chat](https://wordpress.slack.com/archives/C5UNMSU4R/p1525181362000181)

The benefit of being able to lint multiple `package.json` files allows us to ensure specific fields are included in our `package.json` files, for example, `version`, `name`, `description`, and `publishConfig` for Lerna which caught us out recently. `npm-package-json-lint-config` also triggers warnings and errors when fields use incorrect types, e.g. for when an object is expected yet an array is used for a `package.json` value. The configuration also [forces an opinionated sort order](https://github.com/WordPress/packages/pull/119/files#diff-5dc86054e534e818dc9287cb96e2fddaR41) of the `package.json` fields.

The _shared configuration_ is relatively generic which will allow it to be used to validate WordPress projects `package.json` files, for example, [core](https://core.trac.wordpress.org/browser/trunk/package.json), [bbPress](https://bbpress.trac.wordpress.org/browser/trunk/package.json), [BuddyPress](https://buddypress.trac.wordpress.org/browser/trunk/package.json), [Dashicons](https://github.com/WordPress/dashicons/blob/master/package.json), [grunt-patch-wordpress](https://github.com/WordPress/grunt-patch-wordpress/blob/master/package.json), and [Gutenberg](https://github.com/WordPress/gutenberg/blob/master/package.json). For this repo the base configuration of `@wordpress/npm-package-json-lint-config` is _extended_ by a `npmPackageJsonLintConfig ` configuration option in the root `package.json` [here](https://github.com/WordPress/packages/pull/119/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R32).

This PR doesn't remove the manual check for `publishConfig` that was added in #115 as `npm-package-json-lint` currently only allows for checking if `publishConfig` exists, I've opened an issue [here](https://github.com/tclindner/npm-package-json-lint/issues/80) requesting that a `valid-values-publishConfig` rule be added so that we can enforce that the value of `publishConfig` is `"access": "public"`


scripts/create-symlinks.js
```diff
@@ -12,6 +12,7 @@ mkdirp( 'node_modules/@wordpress', () => {
 		[ 'packages/browserslist-config', 'node_modules/@wordpress/browserslist-config' ],
 		[ 'packages/babel-preset-default', 'node_modules/@wordpress/babel-preset-default' ],
 		[ 'packages/jest-preset-default', 'node_modules/@wordpress/jest-preset-default' ],
+		[ 'packages/npm-package-json-lint-config', 'node_modules/@wordpress/npm-package-json-lint-config' ],
```
gziolo: I was wondering if we can create this list using `glob`?

packages/npm-package-json-lint-config/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "@wordpress/npm-package-json-lint-config",
```
aduth: Any way we can enforce tabs for indentation? 😉

packages/npm-package-json-lint-config/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "wordpress",
+    "npm-package-json-lint",
+    "npm-package-json-lint-config"
+  ],
+  "homepage": "https://github.com/WordPress/packages/tree/master/packages/npm-package-json-lint-config/README.md",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/WordPress/packages.git"
+  },
+	"bugs": {
```
aduth: Whitespace consistency.

packages/npm-package-json-lint-config/package.json
ntwb: I've requested a new rule to prevent my future self facepalming:

https://github.com/tclindner/npm-package-json-lint/issues/81

packages/npm-package-json-lint-config/package.json
ntwb: Thanks, fixed in https://github.com/WordPress/packages/pull/119/commits/963eae118d4d124a61472a5c2142e984d5d0c763

package.json
```diff
@@ -56,7 +56,7 @@
 		"check-engines": "check-node-version --package",
 		"create-symlinks": "node ./scripts/create-symlinks.js",
 		"lerna-bootstrap": "lerna bootstrap --hoist",
-		"npmpackagejsonlint": "npmPkgJsonLint ./packages -c .npmpackagejsonlintrc.json",
+		"npmpackagejsonlint": "npmPkgJsonLint ./packages",
```
gziolo: Should it be named `lint-packages`?

scripts/create-symlinks.js
ntwb: Possibly, will follow up in another PR

package.json
ntwb: What about `lint:pj`, not that it is immediately obvious what actually infers I saw it the other day alongside `lint:js`, `lint:md` etc?

```json
		"lint:js": "eslint .",
		"lint:css": "stylelint .",
		"lint:md": "remark-lint .",
		"lint:pj": "npmPkgJsonLint ./packages",
		"lint:scss": "stylelint --syntax=scss .",
```

ntwb: Or `lint:package` maybe?, could be used with `npm-run-all` `npm run all lint:*`

gziolo: `lint:pkg-json` should do the trick, and agreed on `npm-run-all`.

gziolo:
`is-plain-obj` is not added as a dependency, but it works because it is included by something else 😄

gziolo:
I addressed my own comments, let's get it in to take advantage of linting :)

package.json
```diff
@@ -11,8 +11,10 @@
 		"check-node-version": "^3.1.1",
 		"codecov": "^2.3.1",
 		"glob": "^7.1.2",
+		"is-plain-obj": "^1.1.0",
```
aduth: Does this need to be defined as a `devDependencies` in the root if we're not using it in the root?

packages/npm-package-json-lint-config/test/index.test.js
```diff
@@ -0,0 +1,19 @@
...
+import isPlainObj from 'is-plain-obj';
+
+/**
+ * Internal dependencies
+ */
+import config from '../';
+
+describe( 'npm-package-json-lint config tests', () => {
+	it( 'should be an object', () => {
+		expect( isPlainObj( config ) ).toBeTruthy();
```
aduth: The `toBeTruthy` assertion should be discouraged, as it implies that we expect it could be a value other than explicitly `true` (a non-zero number, a non-empty string, etc). If we expect it should be explicitly `true`, we should assert as `toBe( true )`.

packages/npm-package-json-lint-config/test/index.test.js
```diff
@@ -0,0 +1,19 @@
+/**
+ * External dependencies
+ */
+import isPlainObj from 'is-plain-obj';
```
aduth: Minor: Not sure if introduction of dependency here is strictly necessary. What is it we're testing? 

Would [`expect.toMatchObject( {} )`](https://facebook.github.io/jest/docs/en/expect.html#tomatchobjectobject) have been adequate, particularly if we don't care about subset of properties?

Alternatively, Lodash includes an [`_.isPlainObject`](https://lodash.com/docs/4.17.10#isPlainObject) as well, if we want to lean on Lodash as a canonical source of utility functions.

Thinking in terms of being explicit with where dependencies add value to avoid excessive proliferation.

package.json
gziolo: Package are local, tests are global, so we just need pick one approach for those dependencies. I have a feeling that other packages define them in their own `package.json` file. Lerna will hoist, them so we don't need it in here.

aduth: Okay. Seemed we should need it to be defined in _either_ root _or_ `npm-package-json-lint-config/package.json`, but not both.

PR #123, opened 2018-05-07
Fix/wordcount whitespace only error

### LINKED ISSUE #122: WordCount: Error thrown on word count when sanitization results in whitespace-only string
Related: https://github.com/WordPress/gutenberg/issues/6617

When trying to count words for a string which includes only content which would be removed by the sanitization which occurs prior to matching words (HTML tag / comments / etc removal), `@wordpress/wordcount` throws an error:

```
⇒ node          
> var wordcount = require( '@wordpress/wordcount' );
undefined
> wordcount.count( '<p></p>', 'words' );
TypeError: Cannot read property 'length' of null
    at Object.count (/Users/andrew/Documents/Code/vvv/www/editor/htdocs/wp-content/plugins/gutenberg/node_modules/@wordpress/wordcount/build/index.js:121:50)
```

ryanwelcher:
This PR adds logic to ensure that the return of `matchWords()` and `matchCharacters()` is not null before accessing the `.length` property.  A test matching the original issue reported was also added.

Fixes #122


packages/wordcount/src/index.js
```diff
@@ -93,12 +93,13 @@ function matchCharacters( text, regex, settings ) {
 
 export function count( text, type, userSettings ) {
 	const settings = loadSettings( type, userSettings );
+	let returnData = 0;
```
tofumatt: Instead of setting this to `0` and changing it, it might be clearer (and one less line) to just return inside the lower if... eg:

```js
export function count( text, type, userSettings ) {
	const settings = loadSettings( type, userSettings );
	if ( text ) {
		let matchRegExp = settings[ type + 'RegExp' ];
		const results = ( 'words' === settings.type ) ? matchWords( text, matchRegExp, settings ) : matchCharacters( text, matchRegExp, settings );
		if ( results ) {
			return results.length;
		}
		return 0;
	}
}

File: packages/wordcount/src/index.js [hunk 2]
```diff
@@ -93,12 +93,13 @@ function matchCharacters( text, regex, settings ) {
...
 	if ( text ) {
 		let matchRegExp = settings[ type + 'RegExp' ];
-		if ( 'words' === settings.type ) {
-			return matchWords( text, matchRegExp, settings ).length;
-		} else {
-			return matchCharacters( text, matchRegExp, settings ).length;
+		const results = ( 'words' === settings.type ) ? matchWords( text, matchRegExp, settings ) : matchCharacters( text, matchRegExp, settings );
```
tofumatt: This is quite a long line; I get why you're using a ternary here, but it might be nice to newline after the `:` 🤷‍♂️

ryanwelcher:
@tofumatt thank you for your suggestions. I've implemented them with https://github.com/WordPress/packages/pull/123/commits/5e3a2e93f8c5dea6bbf7a79d01b059d91049c2a8.

packages/wordcount/src/index.js
```diff
@@ -95,10 +95,12 @@ export function count( text, type, userSettings ) {
...
 		let matchRegExp = settings[ type + 'RegExp' ];
-		if ( 'words' === settings.type ) {
-			return matchWords( text, matchRegExp, settings ).length;
-		} else {
-			return matchCharacters( text, matchRegExp, settings ).length;
+		const results = ( 'words' === settings.type ) ?
+			matchWords( text, matchRegExp, settings ) :
+			matchCharacters( text, matchRegExp, settings );
+		if ( results ) {
+			return results.length;
```
tofumatt: And I just realised that if we want to go full ternary here you could do:

```js
return results ? results.length : 0;
```

In this case I don't think the ternary hurts readability, so if you want to feel free to do that. 🤷‍♂️

PR #124, opened 2018-05-07
is-shallow-equal: Use implicit index.js for main entry

aduth:
This pull request seeks to resolve an issue with the `@wordpress/is-shallow-equal` module where the `package.json` points to non-existent files for its `main` and `module` entries. `is-shallow-equal` is unique from other packages in that it is not transpiled; its main source lives at `packages/is-shallow-equal/index.js` and can be required directly.

It's not entirely clear to me how this works at all in Gutenberg. I believe it may be that in the [entry point configuration](https://github.com/WordPress/gutenberg/blob/4dc4c7b7b180ee3aef4c7a9acf40d17501d8bc98/webpack.config.js#L126-L130), the `main` field is being ignored or at least falling back to `index.js` where the files do not exist.

In fact, this seems to work fine for Node use as well. It's been reported that the only problematic case with the published version is with React Native. (cc @hypest)

__Testing instructions:__

I found this difficult to test, as via [`npm link`](https://docs.npmjs.com/cli/link) the package was being treated as an ES6 module and reporting an error on assigning to `module.exports`. I was able to have it test correctly by manually updating Webpack to the latest version in Gutenberg, then commenting the [line which considers `is-shallow-equal`](https://github.com/WordPress/gutenberg/blob/4dc4c7b7b180ee3aef4c7a9acf40d17501d8bc98/webpack.config.js#L90) an external module. At which point, `npm run dev` and loading the editor in the browser should produce no errors.


hypest:
Tried manually patching the package locally to match this diff and works fine. 👍 from my side.

PR #128, opened 2018-05-22
Scripts: Add support for lint-pkg-json script

gziolo:
Follow-up for #119.

This PR adds a new script to `@wordpress/scripts`. It follows what we have already for `npm test` and introduces `wp-scripts lint-pkg-json` which helps enforce standards for your package.json file.

This change is required to make it available in Gutenberg, to make sure that packages published to npm follow the same setup as we have in this repository.

### Testing
`npm run lint:pkg-json` - it should work as before. To test it try to change any `package.json` file in the `packages` folder.


packages/scripts/README.md
```diff
@@ -30,6 +30,10 @@ This is how you execute those scripts using the presented setup:
 
 ## Available Scripts
 
+### `wp-scripts lint-pkg-json`
+
+Helps enforce standards for your package.json file. It uses [npm-package-json-lint](https://www.npmjs.com/package/npm-package-json-lint) with the set of default rules provided. You can override them with your own rules as described in [npm-package-json-lint wiki](https://github.com/tclindner/npm-package-json-lint/wiki).
+
```
ntwb: Let's include an example here:

```json
  "scripts": {
    "lint:pkg-json": "wp-scripts lint-pkg-json ."
  },
```

Without the `.` at the end of the command this error is displayed: `No lint targets provided`

gziolo: Yes, good call. We can make sure it is provided as default instead. What do you think?

ntwb: By default is a good idea if it can be overridden using `packages/` for example with Lerna repos.

Regardless of that though we should include some basic docs here in the readme

gziolo: Yes, let's start with your example and iterate :)

gziolo:
> We should also look add more tests in the future

That's always the hardest part with CLI scripts - more mocking than actual testing :D

packages/scripts/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 1.2.0 (2018-05-23)
```
gziolo: I think we will wait for license checker before we release `1.2.0`.

aduth: >I think we will wait for license checker before we release `1.2.0`.

 So will this or won't this be published on `2018-05-23` ? I think we should either:

- Not compile `CHANGELOG.md` edits until immediately before a published release.
   - This is what I thought had been settled upon in a previous core-js meeting ([reference](https://wordpress.slack.com/archives/C5UNMSU4R/p1524577254000794))
- Not merge with an explicit date for a version until the date is known (either omit entirely, or a placeholder value like "Unreleased" or "TBD").

Regardless, we should codify this in `CONTRIBUTING.md` workflow documentation.

gziolo: Yes, `Unreleased` seems like a good choice in this case. I can publish new version just to make sure it doesn't lie 😄

gziolo: For the record, it is still unpublished, we should update docs and follow the flow @aduth described 👍

PR #129, opened 2018-05-22
Add a license check script

pento:
To ensure that any modules we use are GPL2 compatible, we need a script to scan across them and check their licenses.

### DIFF
--- a/packages/scripts/scripts/check-licenses.js
+++ b/packages/scripts/scripts/check-licenses.js
@@ -1,0 +1,119 @@
+// Do this as the first thing so that any code reading it knows the right env.
+process.env.BABEL_ENV = 'development';
+process.env.NODE_ENV = 'development';
+
+// Makes the script crash on unhandled rejections instead of silently
+// ignoring them. In the future, promise rejections that are not handled will
+// terminate the Node.js process with a non-zero exit code.
+process.on('unhandledRejection', err => {
+  throw err;
+});
+
+/**
+ * External dependencies
+ */
+const spawn = require( 'cross-spawn' );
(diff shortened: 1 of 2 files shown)


packages/scripts/scripts/check-licenses.js
```diff
@@ -0,0 +1,119 @@
...
+		return;
+	}
+
+	const filename = path + '/package.json';
+	if ( ! existsSync( filename ) ) {
+		console.error( `Unable to locate package.json in ${path}.` );
+		process.exit( 1 );
+	}
+
+	const package = require( filename );
```
gziolo: `package` is a reserver work, `pkg` isn't as cool, but common in node world.

packages/scripts/scripts/check-licenses.js
```diff
@@ -0,0 +1,119 @@
+// Do this as the first thing so that any code reading it knows the right env.
+process.env.BABEL_ENV = 'development';
```
gziolo: Those 10 lines at the beginning are Jest specific - they can be removed probably.

packages/scripts/scripts/check-licenses.js
ntwb: Did. No. Know. That.

packages/scripts/scripts/check-licenses.js
pento: Ah, cool. I assumed they were necessary, but I'll get rid of them.

ntwb:
Issue moved to [WordPress/gutenberg #7822](https://github.com/WordPress/gutenberg/issues/7822) via [**ZenHub**](https://www.zenhub.com/)

PR #130, opened 2018-05-22
fix: Update allowed licenses, i.e. GPL2 compatible licenses

ntwb:
This PR seeks to update the list of allowed GPL2 compatible licenses based on the list currently included in #129


ntwb:
@pento via https://wordpress.slack.com/archives/C5UNMSU4R/p1526992426000277
> It’s not just GPL compatible licenses.
> w.org allows any OSS license.
> It’s only Core that has the strict GPL2 compat requirement.

So that was the basis of my decision to add more licenses to the allowed list.

A follow up discussion suggested we create a `@wordpress/licenses` package to use as the basic for the list used here and in #129 for the license check package.

An initial pass at the `@wordpress/licenses` package is in #132 

----

In relation to the _Packages_ and _Gutenberg_ repos we can restrict the allowed license to `GPL-2.0-or-later` as is currently done for the respective repos in the `npmPackageJsonLintConfig` config option in the repo [root here](https://github.com/WordPress/packages/blob/master/package.json#L31-L50)

aduth:
Are not these packages considered "Core" ?

ntwb:
Issue moved to [WordPress/gutenberg #7821](https://github.com/WordPress/gutenberg/issues/7821) via [**ZenHub**](https://www.zenhub.com/)

PR #134, opened 2018-06-05
Count when a hook that doesn't have any handlers is run.

### LINKED ISSUE #133: didAction() incorrectly returns 0 until it has a function attached
In Core, `did_action()` returns the number of times a hook has been fired.

In `@wordpress/hooks`, `didAction()` returns the number of times a hook has been fired _after_ a function has been attached to it.

For example, the following test:

```js
const { doAction, addAction, didAction } = require( '@wordpress/hooks' );

function foo() {}

addAction( 'bar', 'foo', foo );

console.log( "We haven't done 'bar' yet", didAction( 'bar' ) );

doAction( 'bar' );

console.log( "Now we have", didAction( 'bar' ) );

console.log( "We haven't done 'baz' yet", didAction( 'baz' ) );

doAction( 'baz' );

console.log( "Now we have", didAction( 'baz' ) );
```

Produces:

```
We haven't done 'bar' yet 0
Now we have 1
We haven't done 'baz' yet 0
Now we have 0
```

However, the same code in PHP:

```php
<?php

include_once( '../wordpress-develop/src/wp-includes/plugin.php' );

function foo() {}

add_action( 'bar', 'foo' );

echo "We haven't done 'bar' yet ", did_action( 'bar' ), "\n";

do_action( 'bar' );

echo "Now we have ", did_action( 'bar' ), "\n";

echo "We haven't done 'baz' yet ", did_action( 'baz' ), "\n";

do_action( 'baz' );

echo "Now we have ", did_action( 'baz' ), "\n";
```

Produces:

```
We haven't done 'bar' yet 0
Now we have 1
We haven't done 'baz' yet 0
Now we have 1
```

pento:
Fixes #133.


packages/hooks/src/test/index.test.js
```diff
@@ -494,6 +494,11 @@ test( 'Test doingAction, didAction and hasAction.', () => {
 	doAction( 'another.action' );
 	expect( doingAction( 'test.action' ) ).toBe( false );
 
+	expect( didAction( 'unattached.action' ) ).toBe( 0 );
+	doAction( 'unattached.action' );
```
kadamwhite: If I duplicate this line (i.e. run the action twice before checking count), didAction still returns 1 where I would expect it to return 2. Am I misunderstanding how this is intended to behave?

pento: You're absolutely correct, good catch, @kadamwhite!

packages/hooks/src/createRunHook.js
```diff
@@ -22,10 +22,17 @@ function createRunHook( hooks, returnFirstArg ) {
 	 */
 	return function runHooks( hookName, ...args ) {
 		let handlers;
-		if ( hooks[ hookName ] ) {
-			handlers = hooks[ hookName ].handlers;
+		if ( ! hooks[ hookName ] ) {
```
aduth: There's an identical condition below in this function which should have been removed as its now redundant (dead code, as the condition will never be satisfied).

PR #135, opened 2018-06-05
Update `CONTRIBUTING.md` with pre-release `CHANGELOG.md` procedure

ntwb:
Updating the `CHANGELOG.md` file before a release ensures a few things:

• The `CHANGELOG.md` file is maintained and up to date as it is included in the package distribution

• Prevents Lerna from triggering a new release of the package because of `CHANGELOG.md` only changes

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -60,6 +60,8 @@ If you're publishing a new package, ensure that its `package.json` file contains
 	}
 ```
 
+Update the packages `CHANGELOG.md` file with the forthcoming version number (`x.y.z`) along with the notable changes included in the release.
+
 ### Development release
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -60,6 +60,8 @@ If you're publishing a new package, ensure that its `package.json` file contains
 	}
 ```
 
+Update the packages `CHANGELOG.md` file with the forthcoming version number (`x.y.z`) along with the notable changes included in the release.
```
gziolo: Can we add a note that it should be put under `Unpublished` section, which should get updated after release gets out? There might be one issue with this flow, updating `CHANGELOG.md` after release will make Lerna think that there needs to be another version bump :P We probably should teach Lerna to ignore this file from its checks.

ntwb:
Issue moved to [WordPress/gutenberg #7817](https://github.com/WordPress/gutenberg/issues/7817) via [**ZenHub**](https://www.zenhub.com/)

PR #136, opened 2018-06-06
Update packages to work with Babel 7

gziolo:
This PR follows the migration docs for Babel 7 migration: https://new.babeljs.io/docs/en/next/v7-migration.html.

It is highly influenced by Calypso migration done a few weeks back on a much larger codebase with the help from Babel maintainers: https://github.com/Automattic/wp-calypso/pull/23424.

The most useful hint:
> - Updated `babel-core` to the [bridge](https://github.com/babel/babel-bridge) which is necessary for anything having babel as a peer dependency -- aka jest.

It is still not working. Some tests are failing. @aduth or @youknowriad , do you have any quick guess what is causing those failures when running `npm run test packages/babel-plugin-makepot/`

I have some issues with `babel-jest` which is no longer working as I would expect. I had to apply `"^.+\\.jsx?$": "<rootDir>/packages/scripts/config/babel-transform.js"` in the main `package.json` to make it work. Ideally, it shouldn't be possible at all when overriding Jest config.


package.json
```diff
@@ -19,14 +19,14 @@
...
-	"babel": {
-		"presets": "@wordpress/default"
-	},
 	"jest": {
 		"collectCoverageFrom": [
 			"packages/*/**/*.js"
 		],
-		"preset": "@wordpress/jest-preset-default"
+		"preset": "@wordpress/jest-preset-default",
+		"transform": {
```
gziolo: We need to find a better way to apply this transform behind the scenes because we will have to apply a similar fix for Gutenberg otherwise.

packages/babel-preset-default/package.json
```diff
@@ -22,12 +22,14 @@
 		"node": ">=8"
 	},
 	"dependencies": {
+		"@babel/plugin-proposal-async-generator-functions": "^7.0.0-beta.49",
+		"@babel/plugin-proposal-object-rest-spread": "^7.0.0-beta.49",
+		"@babel/plugin-transform-react-jsx": "^7.0.0-beta.49",
+		"@babel/plugin-transform-runtime": "^7.0.0-beta.49",
+		"@babel/preset-env": "^7.0.0-beta.49",
+		"@babel/runtime": "^7.0.0-beta.49",
```
gziolo: I'm wondering if we shouldn't apply `runtime` to all packages that have it referenced after Babel does its magic.

packages/jest-preset-default/scripts/pegjs-transform.js
```diff
@@ -1,16 +0,0 @@
-const pegjs = require( 'pegjs' );
```
gziolo: We don't need it outside of Gutenberg.

packages/jest-preset-default/scripts/setup-globals.js
```diff
@@ -13,35 +13,6 @@ global.window.matchMedia = () => ( {
 	removeListener: () => {},
 } );
 
-global.window._wpDateSettings = {
```
gziolo: `@wordpress/date` provides its own defaults after @youknowriad did refactoring.

packages/scripts/package.json
```diff
@@ -34,8 +34,9 @@
 		"@wordpress/babel-preset-default": "^1.3.0",
 		"@wordpress/jest-preset-default": "^1.0.6",
 		"@wordpress/npm-package-json-lint-config": "^1.0.0",
+		"babel-jest": "^23.0.1",
```
gziolo: This can be removed, it's already handled in the preset.

scripts/build.js
```diff
@@ -30,15 +30,15 @@ const ERROR = chalk.reset.inverse.bold.red( ' ERROR ' );
 /**
  * Babel Configuration
  */
-const babelDefaultConfig = require( '../packages/babel-preset-default' );
+const babelDefaultConfig = require( '../packages/babel-preset-default' )();
```
gziolo: Babel expects functions in presets in 7.x line.

packages/jest-preset-default/scripts/pegjs-transform.js
aaronjorbin: Could this have a breaking effect that needs to be documented if someone is using @wordpress/jest-preset-default ?

gziolo: We should include it in the changelog. We can also bump version to 2.0.0 to follow the semantic versioning. I don't think that other code that `blocks` module in Gutenberg would ever need to use it.

gziolo: @nerrad, can you confirm it can be removed or should we keep it as long as `blocks` depend on it?

package.json
```diff
@@ -19,14 +19,14 @@
 		"rimraf": "^2.6.1",
 		"symlink-or-copy": "^1.2.0"
 	},
-	"babel": {
-		"presets": "@wordpress/default"
-	},
```
youknowriad: Why are we removing this, is it useless?

package.json
youknowriad: What's this transform about?

packages/babel-preset-default/index.js
```diff
@@ -1,22 +1,27 @@
...
-	].filter( Boolean ),
+module.exports = function() {
+	return {
+		presets: [
+			! isTestEnv && [ '@babel/preset-env', {
+				modules: false,
+				targets: {
+					browsers: [ 'extends @wordpress/browserslist-config' ],
+				},
+				useBuiltIns: 'usage',
```
youknowriad: If we can make this work, this would be a big improvement ❤️

packages/babel-preset-default/package.json
youknowriad: I'm wondering if we can't just remove the runtime since we're using the polyfills of `babel-env`

packages/jest-preset-default/scripts/pegjs-transform.js
youknowriad: I think this babel7 change needs a bump to `2.0.0` anyway. I'm wondering if we should update the pragma at the same time to avoid the need for a `3.0.0` later

packages/babel-preset-default/package.json
gziolo: Hmmm, `@babel/runtime` and `core-js` are used for requiring code in the transpiled code when using `env` preset. The latter is the dependency of the previous one.

package.json
gziolo: I left a comment in the description:

> I have some issues with `babel-jest` which is no longer working as I would expect. I had to apply `"^.+\\.jsx?$": "<rootDir>/packages/scripts/config/babel-transform.js"` in the main `package.json` to make it work. Ideally, it shouldn't be possible at all when overriding Jest config.

The thing is Jest is still on Babel 6 and we try to make it work with Babel 7, but something is broken and I don't see Babel transforms applied properly.

packages/jest-preset-default/scripts/pegjs-transform.js
gziolo: Hmm, we can flip the pragma handling in Gutenberg, good idea 👍

package.json
gziolo: Jest can read Babel from its config, and the script which generates the distribution folders for packages uses the same preset internally. So it isn't necessary anymore - it uses defaults :)

packages/jest-preset-default/scripts/pegjs-transform.js
nerrad: I still don't totally grok what pegjs is used for (Does it provide the grammar rules for the custom html comments for blocks?).  Assuming that's what its for, and it only is required by GB `blocks` - plugins writing tests can probably just mock any GB blocks in their tests.

packages/babel-preset-default/index.js
```diff
@@ -1,22 +1,27 @@
...
+			} ],
+			isTestEnv && [ '@babel/preset-env', {
+				useBuiltIns: 'usage',
+			} ],
+		].filter( Boolean ),
+		plugins: [
+			'@babel/plugin-proposal-object-rest-spread',
```
aduth: Not sure if the Babel core has been updated for it yet, but Rest/Spread was ratified as part of ES2018, so it might be possible to remove this.

https://github.com/tc39/proposals/blob/master/finished-proposals.md

package.json
gziolo: I had to add `babel.config.js` instead to make everything work 😄

scripts/build.js
gziolo: Now I see:

> const isTestEnv = api.env() === 'test';
> TypeError: Cannot read property 'env' of undefined

because we need to pass `api` object to this preset ...

ntwb:
Closing this per @gziolo recommendation (work to begin shortly on this directly in the Gutenberg repo)

PR #137, opened 2018-06-08
Jest Console: Add new matchers for console.log and console.info

gziolo:
It was raised a few times in the past that `@wordpress/jest-console` supports only `console.error` and `console.warn`. We discussed that we should have a similar handling for `console.log` and `console.info`. This PR adds 4 new methods to `@wordpress/jest-console` package:

### `.toHaveInformed()`

Use `.toHaveInformed` to ensure that `console.info` function was called.

Almost identical usage as `.toHaveErrored()`.

### `.toHaveInformedWith( arg1, arg2, ... )`

Use `.toHaveInformedWith` to ensure that `console.info` function was called with
specific arguments.

Almost identical usage as `.toHaveErroredWith()`.

### `.toHaveLogged()`

Use `.toHaveLogged` to ensure that `console.log` function was called.

Almost identical usage as `.toHaveErrored()`.

### `.toHaveLoggedWith( arg1, arg2, ... )`

Use `.toHaveLoggedWith` to ensure that `console.log` function was called with
specific arguments.

Almost identical usage as `.toHaveErroredWith()`.

## Testing
`npm test` should pass.

You can also add `console.log` and `console.info` in the source code or unit test to make sure it fails test that is using it when the matcher is not provided.
```js
console.log( 'Test will fail' );
```

In addition, you can add expectation with matcher that is not used in the code to make sure that it fails because condition was not met:
```js
expect( console ).toHaveLogged();
```


packages/jest-console/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## Unreleased (2.0.0)
```
gziolo: It will also make `wp-scripts test-unit-js` fail when code uses `console.log` and `console.info`, so we will have to bump `@wordpress/scripts` - ideally we should align it with Babel 7 upgrade.

packages/jest-console/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
...
+
+- Add new API methods `toHaveInformed`, `toHaveInformedWith`, `toHaveLogged` and `toHaveLoggedWith` ([137](https://github.com/WordPress/packages/pull/137))
```
aduth: We should clarify that this is a breaking change ("Breaking:" prefix) and some hint to in what way it breaks / how consumers can update their code.

gziolo: Good point, I will add details.

